PR #4, opened 2016-02-24
Harmony app proto-prototype

jeffbaumes:
This also does some reorg of the main library which still needs some work.


package.json
```diff
@@ -4,7 +4,8 @@
   "description": "Visualization components for the web",
   "main": "index.js",
   "scripts": {
-    "build": "webpack"
+    "build": "webpack",
+    "build-harmony": "cd app/harmony && webpack"
```
ronichoudhury: Can we use ``build:harmony`` instead? I feel like the colon "scopes" the build (or other) subcommands better. (If we're ok with the change, it needs to be changed in the Harmony README as well.)

ronichoudhury:
Aside from the line item above, LGTM.

I'm going to reorg the demo app in a separate PR.

jeffbaumes:
I've updated the script name to use colon.

ronichoudhury:
If you update the README as well I think this is ready to merge.

jeffbaumes:
Done.

ronichoudhury:
Should we follow the "PR owner presses the launch button" protocol?

jeffbaumes:
Yes please, but I'm looking for a LGTM - "i think this is ready to merge" is not official enough for my taste.

ronichoudhury:
Oh, I thought my earlier "LGTM" would suffice :)

*ahem*... LGTM!

jeffbaumes:
Hooray! You are right you did give a provisional LGTM that should have sufficed.

PR #8, opened 2016-02-25
Resonant reference app

jeffbaumes:


jeffbaumes:
@alex-r-bigelow FYI I'm going to put a commit on this branch to update the readme to reflect the app's change as a Girder plugin.

jeffbaumes:
I think this is ready to go as an initial version.

jeffbaumes:
Oh @alex-r-bigelow do you want to refactor the dir structure first? I'll leave this until tomorrow.

alex-r-bigelow:
That's kind of funny... I was just writing stuff down in the README as well to help me remember the build process. You can go ahead, and I'll add my comments later

app/resonant-reference-app/core_models/dataset.js
```diff
@@ -0,0 +1,53 @@
...
+        // TODO
+    },
+    getMetadata : function () {
+        // TODO
+    }
+});
+
+Dataset.DATA_TYPES = DATA_TYPES;
+Dataset.LOCATION_TYPES = LOCATION_TYPES;
+module.exports = Dataset;
```
jeffbaumes: I see some files with no EOL at the end of the file. Can you check if your editor has an option to ensure this on save? Even if not, can you fix these :) ?

jeffbaumes:
The organization looks much better. Can I ask you to do even a bit more? Here are my suggestions:

* Move code from `tools` and `core_views` into a single `views` dir.
* Rename `core_models` to just `models`.
* Rename `css_extras` to `stylesheets`.

alex-r-bigelow:
I feel like the separation between `tools` (I agree I should at least rename this to `tool_views`) and `core_views` is kind of important... `core_views` manage the reference app interface itself and they're somewhat inter-dependent, but `tool_views` are a bit more modular—they're just embedded inside the interface, and could be embedded elsewhere.

But I'm cool with flattening even further if that's less trouble / conforms to Kitware coding standards (are these written down anywhere?)

jeffbaumes:
Then lets go with something like `views/layout` and `views/widgets` instead of using underscores. Each should have a dir per view (because each view has its template alongside it which I think is nice). Use these subdirs as inspiration:

https://github.com/Kitware/minerva/tree/master/web_external/js/views

jeffbaumes:
I basically want Kitware web apps to use common structure when possible. There is no set standard but it would ease developers to not entirely switch mental models when switching apps.

alex-r-bigelow:
I'm just about finished with the restructuring, as well as #9 and #16, but I'm running into the telegraph errors. @ronichoudhury, how does one get around this, again?

jeffbaumes:
I believe you comment out lines 1 and 11 from https://github.com/Kitware/candela/blob/master/src/core/VisualizationComponent.js.

alex-r-bigelow:
I've pushed the reshaping + code style changes (#9 and #16)... I'm still getting odd upstream errors, so for now I commented out the references to candela.js

ronichoudhury:
@alex-r-bigelow I'm going to take a look at the errors you're getting today. Can you give me a little bit more background on what you're seeing?

alex-r-bigelow:
When I import (or require) candela.js, I get a

    Cannot read property 'TYPED_ARRAY_SUPPORT' of undefined

For now, I've uncommented line 3 of views/widgets/SingleVisualizationView/SingleVisualizationView.js

It looks like it's part of some deep library that candela pulls in—it's possible I've broken my candela build somehow. I'm working on nuking everything and starting fresh right now; I'll let you know if I figure anything out on that end.

jeffbaumes:
Layout structure, capitalization, etc. is looking really good. Once we sort out this error it should be about ready to merge.

jeffbaumes:
I can also see @alex-r-bigelow's error when uncommenting that line and commenting out the telegraph stuff.

alex-r-bigelow:
I just jumped back to 04b26416ec4d27121d45b7e397dc4f8f121304f0 to triple-check my candela build, and everything works fine—it imports stuff and runs the visualization without any problems. As the only thing that really changed since that commit other than whitespace was switching over to ES6 syntax, I'm guessing that has something to do with it.

Remind me why we `require()` candela instead of `import`? I'm still new to this ES6 stuff, and I don't have a good feel for the difference between the two...

jeffbaumes:
I think I just fixed the candela import - it uses import now instead of require.

alex-r-bigelow:
Yup, that worked! Thanks!

ronichoudhury:
@jeffbaumes @alex-r-bigelow I'm not able to reproduce the telegraph problem. I'm heading home for the day, but Alex, tomorrow maybe I can have you demonstrate it to me.

alex-r-bigelow:
Sorry about that—apparently there were fires to put out on the infovis paper.

It looks like Jeff's fix also dealt with that—I just rebuilt candela with the telegraph stuff reenabled, and things are working just fine.

-Alex

jeffbaumes:
Huh, you are right, it is fixed. Inadvertent win! Ok, I've removed the WIP tag and the app structure looks good. LGTM. @alex-r-bigelow you may merge and delete the branch.

PR #18, opened 2016-02-29
Minimal preparation for publishing an NPM package

ronichoudhury:
This is to lay claim to the NPM package ``candela`` which was recently awarded to me due to a squatter sitting on the name previously.

@jeffbaumes please take a look. I just need a basic sanity check - later we can make changes to how we structure the code for publishing.


jeffbaumes:
Sanity check passes. LGTM.

PR #22, opened 2016-03-10
Adding initial resize capability and manual demo for testing

jeffbaumes:


ronichoudhury:
I noticed that the linter wasn't being run on stuff in ``src`` without updating the semistandard preloader rule in the webpack config, but I pushed a commit for that.

If that LGTY, then it LGTM :)

jeffbaumes:
Yes that is fine, the code as it was originally did not pass my editor linting check (two `include` keys). Your fix looks right/better.

PR #23, opened 2016-03-10
Girderify

alex-r-bigelow:
The mapping interface appears to be working—if you guys get a minute, could I get you to try it?


app/resonant-reference-app/models/Toolchain.js
```diff
@@ -1,52 +1,177 @@
...
-});
+import Dataset from './Dataset';
+let girder = window.girder;
 
 /*
     A Toolchain represents a user's saved session;
     it includes specific datasets, with specific
-    matchings to specific visualizations (in the future,
-    this may also include faceting settings).
+    mappingss to specific visualizations (in the future,
```
jeffbaumes: Extra `s` here

jeffbaumes:
This is looking very good code-wise and when I look at the application. I think there are some UI simplification steps to think about (I had to click around before I started to get the hang of it), but I'm impressed how many things work, even going back and changing the data, expanding all the panels, etc. I don't think this PR is supposed to address all UI issues, so LGTM.

I see that the toolchain has multiple datasets and multiple visualizations (they are arrays). Since we will be doing single-data, single-vis toolchains for a while, I don't want support for multiple data/vis to add too much complication, but I don't think it does so it's ok to leave it in.

alex-r-bigelow:
Yeah, it's definitely not super polished yet, but I think it's almost at the walking skeleton level. I mainly left the arrays (multiple datasets + multiple visualizations) in so that we wouldn't have to completely rewrite the mapping interface again when we get there... for now, it's pretty trivial to just grab the first element.

alex-r-bigelow:
Other than #26, I think the rest of the changes are mostly aesthetic at this point.

I totally agree with #29 --- I need to do a redesign anyway to fit the look and feel of the app a little better, and I can address the usability cues along the way. The critical bits are there now... per your earlier LGTM I'll go ahead and merge.

jeffbaumes:
@alex-r-bigelow thanks for all the changes! In the future, a LGTM only applies to the current state of the PR. If you push some commits, you'll need another LGTM. In this case, I needed to go back and verify that you resolved each of the issues that you fixed on the rest of the branch, but that should have been done before a merge. I was assuming the issues I opened would be resolved in future PRs. No biggie.

alex-r-bigelow:
Oops, sorry about that! Still kind of new to doing real collaborative stuff on github...

PR #30, opened 2016-03-16
Added new scatterplot to vcharts with more options

jeffbaumes:
Added some better defaults for colors and new size and shape scales to make it more interesting. Here are the features working with the reference app:

screen shot 2016-03-15 at 10 36 31 pm


jeffbaumes:
@alex-r-bigelow can you confirm that this doesn't have any major problems and give a LGTM if it seems ok?

alex-r-bigelow:
LGTM! Just tried it again and everything seems fine.

PR #31, opened 2016-03-16
Aesthetics

alex-r-bigelow:
I think I'm going to start some drastic changes (Girder integrations, data editing apps and such), and this is probably a good point to sync up.


jeffbaumes:
When I change the type of a field manually, the things it's allowed to connect to change appropriately, but the type displayed under the field in the mapping panel does not get updated.

jeffbaumes:
I've verified that the unset operation now works as expected.

Scroll for medium-sized number of fields (~20) looks ok, though I might expect the vis attributes to stay in the view at all times while the field list scrolls. This can be addressed in a future PR.

jeffbaumes:
@alex-r-bigelow Please comment on the PR once you've resolved the field type update issue - should be ready to merge at that point but I'll take a final look.

alex-r-bigelow:
I've fixed the type field issue. A local test looks like our branches play well together, so it looks good to me.

jeffbaumes:
One more thing. You should remove mapping links when a new manually specified type does not match a supported type in the vis attribute. Currently you can be in a state where bad links remain by performing a mapping then updating the field's type.

alex-r-bigelow:
Good catch! It should be fixed now.

jeffbaumes:
LGTM! Please go ahead and merge the PR.

PR #33, opened 2016-03-18
Histogram view

jeffbaumes:


src/external/vcharts/src/index.js
```diff
@@ -361,8 +374,14 @@ let chart = function (type, initialOptions, done) {
     // Transform pass 2 to get the final visualization
     that.spec = transform(that.template, curOptions);
 
+    console.log(that.spec);
```
ronichoudhury: Can we get rid of this ``console.log()`` statement?

src/external/vcharts/src/index.js
```diff
@@ -361,8 +374,14 @@ let chart = function (type, initialOptions, done) {
...
+
     vg.parse.spec(that.spec, function (chartObj) {
       let chart = chartObj(vegaOptions);
+      let printSignal = (name, val) => console.log(name + ': ' + val);
+      for (let i = 0; i < that.spec.signals.length; i += 1) {
+        chart.onSignal(that.spec.signals[i].name, printSignal);
+      }
```
ronichoudhury: Same question here I guess, unless the log statements are intentional.

jeffbaumes: No, this is unintentional debug code. Will fix.

src/external/vcharts/src/index.js
jeffbaumes: This is unintentional debug code. Will fix.

alex-r-bigelow:
I like that there's support for strings. There's some odd behavior with really big margins when there are a lot of categories (this is the cars dataset):

screen shot 2016-03-18 at 1 24 24 pm

... but that's really minor. Outside the console stuff, it LGTM.

ronichoudhury:
LGTM too.

I'm intending to create an image comparison testing infrastructure which would be used to add tests for components such as these.

jeffbaumes:
@ronichoudhury I've fixed the console output.

@alex-r-bigelow I think that is due to rounding the bin sizes. If there are 1px bins things like that happen. If there are even more bins I've seen the whole thing disappear with just one line in the center.

I also snuck in axis titles. This is not controversial (hopefully) so I'll go ahead and merge.

src/components/Histogram.js
```diff
@@ -12,7 +12,13 @@ export default class Histogram {
     this.chart = vcharts.chart('histogram', {
       el: el,
       values: data,
-      bin: options.bin
+      bin: options.bin,
+      xAxis: {
+        title: options.bin
+      },
+      yAxis: {
+        title: 'Count'
```
ronichoudhury: Generally speaking, we might want to make this configurable. For instance, a particular user of this histogram may want to specify the units of counting in the label, etc.

ronichoudhury: (I opened #34 to reflect this.)

PR #37, opened 2016-03-21
Add aggregation option to histogram

### LINKED ISSUE #35: Create aggregate by property for histogram
Instead of counting each record as 1, enable an optional field for summing up the bin value

jeffbaumes:
Fixes #35

### DIFF
--- a/app/resonant-reference-app/views/overlays/VisualizationLibrary/index.js
+++ b/app/resonant-reference-app/views/overlays/VisualizationLibrary/index.js
@@ -48,6 +48,10 @@ let VisualizationLibrary = Backbone.View.extend({
         {
           name: 'bin',
           type: 'string'
+        },
+        {
+          name: 'aggregate',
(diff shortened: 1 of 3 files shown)


ronichoudhury:
LGTM.

Is there generally a way to write unit tests for this sort of behavior?

alex-r-bigelow:
LGTM. Very cool!

jeffbaumes:
@ronichoudhury Yes there are a few options. 1. we could do image testing, 2. we could do SVG rendering and ensure some elements (like rects) are the right size, 3. we could ensure that the produced Vega spec is what we expected. I think the third option would be easiest to implement.

ronichoudhury:
Option 3 is closest to what I think of as "unit testing". Option 1 is probably the easier "image test" to perform. I'll look into adding both test types once the testing infrastructure is a bit more bulletproof.

PR #39, opened 2016-03-22
Cleaner output from testing command

ronichoudhury:
This adds a command line filter to the NPM test script (along with some testing configuration changes) to capture and reformat the output in such a way that ``faucet`` is better able to process and pretty print the results.

### DIFF
--- a/config/karma-base.conf.js
+++ b/config/karma-base.conf.js
@@ -2,7 +2,7 @@ module.exports = {
   basePath: '..',
   singleRun: true,
   client: {
-    captureConsole: false
+    captureConsole: true
   },
   browsers: [
     'PhantomJS'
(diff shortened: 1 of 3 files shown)


jeffbaumes:
I cannot see a difference compared to the `master` or `reorg` branches (maybe because we are not using phantom yet?). But LGTM regardless - I'm assuming it will help in that case.

ronichoudhury:
Hm. You may see a difference when a test fails...

All this does is extract the pure tap output from Phantom, then run that through faucet.

ronichoudhury:
I added another variation of ``npm test`` that doesn't process the output at all (as a fallback in case something weird happens).

I'm assuming the LGTM still stands.

ronichoudhury:
Thanks! LGTM.

jeffbaumes:
Raw test output task looks good.

PR #40, opened 2016-03-23
Reorganize source code

ronichoudhury:
The main change here is to put candela and vcharts into their own top-level directories. Other components of this repo (currently, Clique) will be imported similarly.

Further changes, such as integrating vcharts more fully into the build process, will come in later PRs.


jeffbaumes:
LGTM

PR #41, opened 2016-03-23
Production webpack configs

ronichoudhury:
This adds two webpack config files for creating unminified and minified versions of the Candela library for distribution.


ronichoudhury:
We might want to cook up and document a procedure for publishing NPM releases as part of this PR as well. We could create a 0.3.0 release to test that out. Let me know what you think @jeffbaumes

jeffbaumes:
LGTM. The only thing I want to capture here is that we should also test these build products to ensure they work as expected from the point of view of an external product using them. +1 to documenting the release process.

ronichoudhury:
+1 to testing as well.

I will open new issues for testing the dist products, and for documenting the release process.

PR #47, opened 2016-03-24
Allow multiple fields in reference application

### LINKED ISSUE #43: Enable "list of fields" option in reference app
This could be enabled with any special flag for now, but eventually should support the "string_list" type specified in the Candela API doc

jeffbaumes:
Fixes #43.

Add an allowMultiple flag for now in the reference app to support this.

Update BoxPlot's fields argument to support multiple fields


app/resonant-reference-app/models/Toolchain.js
```diff
@@ -132,16 +132,33 @@ let Toolchain = girder.models.ItemModel.extend({
     let self = this;
     let meta = self.get('meta');
     let options = {};
-    
+
+    // Figure out which options allow multiple fields
+    meta.mappings.forEach((mapping) => {
```
ronichoudhury: If you like you can omit the parens here (in case you didn't know).

Unless there is a prevailing opinion that such parens should always be kept (similar to the C/C++ style rule about always keeping curly brackets in loops/conditionals).

jeffbaumes:
@ronichoudhury Good catch. I purged all of these single-arg function parens from the repo, searching for the regex `\([A-Za-z_][A-Za-z0-9_]*\) =>`. Would be nice if semistandard checked for that.

ronichoudhury:
Seems to work very nicely.

LGTM.

jeffbaumes:
Did a little cleanup on the box plot so I could test it, but otherwise the merge went ok. Merging.

PR #48, opened 2016-03-24
Implement real options specs

jeffbaumes:
Depends on #47.

 - The reference app now correctly queries from candela specs.
 - Works for vis where the spec field is defined.
 - Switches allowMultiple to string_list type checking as per spec.
 - Updates specs for BoxPlot, Scatter, Histogram.
 - Updates mapping view logic to conform to spec.


jeffbaumes:
@alex-r-bigelow Note that in the mapping view the `type` field for nodes now has two interpretations. For data fields, it is a single type. For vis options, it is a list of allowed types. The `Dataset.COMPATIBLE_TYPES` mapping is hardly used now, opting for an explicit list of supported types in each vis option as we had discussed.

ronichoudhury:
LGTM

PR #50, opened 2016-03-24
Scatterplot matrix

jeffbaumes:
Fixes #49. Depends on #47 and #48.

PR #57, opened 2016-03-25
Lineup

jeffbaumes:
A LineUpJS component.

Depends on #47 #48 #50. Look at the last commit for this topic's changes.


ronichoudhury:
Very cool. LGTM.

ronichoudhury:
Oh, wait! DLGTMJY so hang on...

ronichoudhury:
I added a small commit to fix the import statements that we talked about earlier today. No big deal but I like to have all node_modules imports look the same.

If that LGTY, then it LGTM.

ronichoudhury:
Actually, it looks like they have fixed their package.json's main property on master, so I think we can just leave our code as is (since we rely on a fixed hash checkout).

PR #61, opened 2016-03-26
Fixup Bar options

jeffbaumes:
It's now able to be used in the reference app.


ronichoudhury:
What happens to the color and hover options?

jeffbaumes:
Discussed with @ronichoudhury IRL and addressed his question. Basically color and hover are not exposable as field options, but should be and then made part of the options spec. I'll make an issue for this.

PR #62, opened 2016-03-26
Create separate vcharts el option

jeffbaumes:
This makes the vcharts api closer to candela's.

Depends on #61.


ronichoudhury:
This seems ok to me but it's hard to judge because I don't fully have a handle on our options spec API just yet. It may be time to start putting together some documentation.

In the meantime, I'll say LGTM if it LGTY.

PR #69, opened 2016-03-28
Bar color hover

### LINKED ISSUE #65: Fixup bar color and hover options
These options should enable fields to be set to them instead of static color properties.

jeffbaumes:
Bar chart color and hover options are now implemented like other plots and are available in the reference app and example code. Fixes #65.

### DIFF
--- a/app/examples/visualizations.json
+++ b/app/examples/visualizations.json
@@ -39,11 +39,19 @@
     "description": "A simple bar chart",
     "component": "Bar",
     "options": {
-      "data": "iris",
-      "x": "petalLength",
-      "y": "petalWidth",
-      "color": "steelblue",
-      "hover": "green"
+      "data": [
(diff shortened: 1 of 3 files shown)


ronichoudhury:
This LGTM, but is there a reason not to use the iris data to drive the example (by using, e.g., species name)?

jeffbaumes:
You cannot make a true bar chart unless you have field with a unique value for each row to use as the x value. Otherwise some bars will overlap with each other. The iris data has no such field.

jeffbaumes:
I've thought about enforcing this as a field attribute ("unique" or something) and restricting the x value to that - but I suppose there might be cases where you want the overlapping bar behavior. We could also always provide an "index" field that is the column number to use as a unique field whenever one is needed. For LineUp I already needed to construct such a field.

PR #71, opened 2016-03-28
Enable comments in JSON

### LINKED ISSUE #36: Enable comments when loading JSON
Perhaps use [this](https://github.com/brandon-arnold/strip-json-comments-loader).

jeffbaumes:
Fixes #36.

### DIFF
--- a/src/candela/components/Scatter/spec.json
+++ b/src/candela/components/Scatter/spec.json
@@ -4,11 +4,19 @@
     ["width", 800],
     ["height", 500],
     ["legend", true],
-    ["hover", [["@get", "x"], ["@get", "y"]]]
-  ],
+
+    // Default to showing the x and y fields on hover.
+    ["hover", [["@get", "x"], ["@get", "y"]]],
+
+    // Default axis titles to x and y field names.
+    ["xAxis.title", ["@get", "x"]],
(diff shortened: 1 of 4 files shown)


src/candela/components/Scatter/spec.json
```diff
@@ -22,7 +30,6 @@
           {
             "axis": "x",
             "size": ["@get", "width"],
-            "title": ["@get", "xAxis.title", ["@get", "x", ""]],
```
ronichoudhury: Was this deletion left over from the options fixup?

src/candela/components/Scatter/spec.json
```diff
@@ -45,7 +52,6 @@
           {
             "axis": "y",
             "size": ["@get", "height"],
-            "title": ["@get", "yAxis.title", ["@get", "y", ""]],
```
ronichoudhury: Same question as above.

ronichoudhury:
LGTM.

src/candela/components/Scatter/spec.json
jeffbaumes: See the diff lines 11-13 above. Instead of putting the default values inline, I just moved them to the top, which I think is better practice and is the point of the `defaults` command.

PR #72, opened 2016-03-28
Some scatterplot matrix fixes

jeffbaumes:
- Fixed bug where vis did not render when color was unspecified
 - Orienting x and y axis labels better
 - Leaving more room for the legend


ronichoudhury:
LGTM

PR #74, opened 2016-03-29
Cleanup vcharts

### LINKED ISSUE #54: Figure out how to get axis.json out of vcharts
``axis.json`` is a special "sub-specification" that many of the other specs make use of. As such, it is probably more properly a "mixin" and we should look into how to treat it as such within the candela components that use it.

jeffbaumes:
Fixes #54. Getting rid of `vcharts` dir and moving main code over to `util/vega`.


ronichoudhury:
LGTM.

We will need to "dissolve" the README.md file into whatever documentation system we end up using. I'll open an issue for it.

jeffbaumes:
Indeed. Thanks for opening the issue.

PR #76, opened 2016-03-30
Coerce datatypes in the reference app when piping data to the visualization

alex-r-bigelow:
Simple test: try connecting Year (cars dataset) to color (with a scatterplot), and switch between string/number/integer/date

However, it looks like a lot of the views don't actually expect Date objects...

### DIFF
--- a/app/resonant-reference-app/models/Toolchain.js
+++ b/app/resonant-reference-app/models/Toolchain.js
@@ -64,7 +64,7 @@ let Toolchain = girder.models.ItemModel.extend({
 
     let meta = self.get('meta');
     if (newDataset.exampleToolchain &&
-        meta.visualizations.length === 0) {
+      meta.visualizations.length === 0) {
       // The user is starting off with this dataset;
       // we want to load up the example visualization and
       // the mappings that goes with it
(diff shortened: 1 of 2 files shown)


jeffbaumes:
Thanks @alex-r-bigelow. I'll cherry-pick this commit on what I'm doing to see what happens. If it works we can merge this then I'll rebase my work on top.

app/resonant-reference-app/models/Toolchain.js
```diff
@@ -134,7 +167,7 @@ let Toolchain = girder.models.ItemModel.extend({
     let options = {};
 
     // Figure out which options allow multiple fields
-    meta.mappings.forEach(mapping => {
+    meta.mappings.forEach((mapping) => {
```
jeffbaumes: I think I went through and removed these parentheses at some point - do you think we should keep them or ditch them as a styling rule?

jeffbaumes:
Oddly, this patch ended up not helping for supporting time series (#82), because Vega requires time to be supplied as millisecond integers. So for now I suggest we close down this PR, though let's keep the code in this branch around for a bit in case we come back to it.

PR #78, opened 2016-03-30
Introduces mixins, including an Events mixin

ronichoudhury:


ronichoudhury:
@jeffbaumes could you look this PR over?

PR #79, opened 2016-03-30
Resize mixin

ronichoudhury:
Adds a mixin to handle autoresizing of containing element. The mixin emits a ``resize`` event, allowing each component to react properly to the notice. For example, the example resize application simply prints a message on the console and then calls ``render()`` to refresh the component.

Depends on #78


ronichoudhury:
Closing in favor of #80

PR #82, opened 2016-04-01
Adding support for time axes and type inference

jeffbaumes:
- Use datalib for type inference which places a `__types__` field on the data input.

 - If the `__types__` is there, use it to infer axis types. Otherwise infer types when initializing the Vega chart.


jeffbaumes:
I've added an example to show type inferencing and date scales ("Date Histogram") but I should also add a few tests for the vega util code that has changed. I'll do that now.

jeffbaumes:
Ok tests are added. Ready for review.

src/candela/util/vega/test/transform.js
```diff
@@ -9,6 +9,9 @@ test('vcharts.transform() without spec', t => {
   t.deepEqual(true, vcharts.transform(true), 'boolean transform should be identity');
   t.deepEqual(null, vcharts.transform(null), 'null transform should be identity');
 
+  let d = new Date();
```
ronichoudhury: I don't think it makes a huge difference but I'd feel safer if this constructed a particular date, rather than using the current time to make one (i.e., the testing state shouldn't depend on the current time, or any other external state).

src/candela/util/vega/test/transform.js
```diff
@@ -9,6 +9,9 @@ test('vcharts.transform() without spec', t => {
...
+  t.deepEqual(d, vcharts.transform(d), 'Date objects should be identity');
```
ronichoudhury: Can you change the descriptor to ``Date object transform should be identity``?

src/candela/components/Line/spec.json
```diff
@@ -0,0 +1,212 @@
...
+              //               },
+              //               {"value": 1}
+              //             ]
+              //           }
+              //         }
+              //       }
+              //     }
+              //   ]
+              // }
+            ]
```
ronichoudhury: Can this commented-out section be removed?

ronichoudhury:
This looks great. I just had the few concerns in the line notes above. After we resolve those I can stamp this.

src/candela/util/vega/test/transform.js
jeffbaumes: Done.

src/candela/util/vega/test/transform.js
jeffbaumes: Done.

src/candela/components/Line/spec.json
jeffbaumes: Done.

jeffbaumes:
@ronichoudhury thanks for the review, your comments made me recheck a few things. There were some missing transform functions that I needed to add (greater than and length) to get the legend logic correct. Should be ready to go now. You can actually check the vis with the line example now.

ronichoudhury:
LGTM!

PR #83, opened 2016-04-01
Auto resize mixin rebased

jeffbaumes:
This is a copy of #81 rebased on current master so I can have a look at the proper diff.

### DIFF
--- a/src/candela/VisComponent/mixin/AutoResize.js
+++ b/src/candela/VisComponent/mixin/AutoResize.js
@@ -1,0 +1,18 @@
+import 'javascript-detect-element-resize/detect-element-resize';
+import Events from './Events';
+
+let AutoResize = Base => class extends Events(Base) {
+  constructor (...args) {
+    super(...args);
+
+    window.addResizeListener(this.el, () => {
+      const style = window.getComputedStyle(this.el);
+      const width = window.parseInt(style.getPropertyValue('width'));
+      const height = window.parseInt(style.getPropertyValue('height'));
+
+      this.emit('resize', width, height, this);
+    });
+  }
+};
+
+export default AutoResize;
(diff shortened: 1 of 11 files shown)


src/candela/VisComponent/mixin/AutoResize.js
```diff
@@ -0,0 +1,16 @@
+import Resize from './Resize';
+
+let AutoResize = Base => class extends Resize(Base) {
```
jeffbaumes: I feel like AutoResize should extend InitSize, since I can't think of where you would want AutoResize and not AutoSize.

jeffbaumes:
LGTM. @ronichoudhury: up to you if you want AutoResize to include InitSize (I can't think of a case where you'd want AutoResize and not InitSize). You can merge this (pre-rebased) and close your other 3 PRs or merge your other 3 PRs and close this one.

src/candela/VisComponent/mixin/AutoResize.js
ronichoudhury: I agree. I think I may have left it out because I've observed that the ``Resize`` mixin causes a resize event to be fired as soon as the element is initialized, which for ``AutoResize`` components will sort of repeat that initial work. But (1) I am not certain that it is guaranteed that a resize event will fire and (2) the tiny bit of extra initialization overhead is probably worth making it explicit that ``AutoResize`` works off the conceptual base provided by ``InitSize``.

jeffbaumes: I agree with (2). If there is reinitialization, we could file a separate optimization issue to be resolved later.

ronichoudhury:
Closing in favor of #78, #80, and #81.

PR #88, opened 2016-04-05
Fix mixin name

jeffbaumes:
Simple typo that style checking found.


ronichoudhury:
Grr. Looks like I didn't run the build before I opened that PR. :smirk_cat:

LGTM.

PR #90, opened 2016-04-05
Fix export for ES5

jeffbaumes:
See https://github.com/webpack/webpack/issues/706 for details.


ronichoudhury:
LGTM.

We should possibly write a test to repeat the exports-structure stuff for the built library file.

PR #95, opened 2016-04-06
Heatmap

jeffbaumes:
A basic heatmap - there is an example and it should work with the reference app too.


ronichoudhury:
Cool! LGTM.

PR #96, opened 2016-04-07
Let the reference app be built by the main Girder build

jeffbaumes:
Adapted from https://github.com/osumo/osumo/blob/master/Gruntfile.js


alex-r-bigelow:
Nice! This makes more sense than the `export GIRDER_PATH=...` nonsense I was doing the other way (running grunt from webpack). LGTM

jeffbaumes:
Merging - there may be bugs since I haven't used it much yet but we'll work those out later.

PR #97, opened 2016-04-07
Fix girder build

jeffbaumes:
Turns out my previous PR didn't work :) This fixes up a few things and make the plugin work a bit more like the osumo plugin by serving a directory, which lets us split out the index.html file.

### DIFF
--- a/app/resonant-reference-app/.gitignore
+++ b/app/resonant-reference-app/.gitignore
@@ -1,2 +1,2 @@
-web_client/extra
+web_client/lib
 *.pyc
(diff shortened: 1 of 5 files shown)


ronichoudhury:
@alex-r-bigelow can you take a look at this?

jeffbaumes:
Forgot this was a dependency of #100 - it got merged just now. @alex-r-bigelow please check out this functionality regardless.

alex-r-bigelow:
Yes, `grunt` by itself seems to work, but `grunt watch` doesn't seem to notice any changes (even if you force a webpack build while it's watching). I imagine it's not supposed to?

jeffbaumes:
Right I did not work to get `grunt watch` working. Glad it is working for you.

PR #98, opened 2016-04-08
Consolidate all examples into single app infrastructure

ronichoudhury:
This PR adds the following capabilities:
- Examples for components can now specify ``elementType`` to use an element besides a ``div``
- Arbitrary app code can now live in the "custom" subdirectory of "app/examples", with an entry in ``visualizations.json`` that specifies ``template`` rather than ``component`` etc.

There are some statefulness issues involving using the back button to back out of a custom app, but these should not impact casual example browsing, nor image-based testing.


ronichoudhury:
Since it's mainly just us writing the custom example code, I'm ok with enforcing a bit more structure in those examples to make it easier to track window listeners and intervals etc. I may just do that in a separate PR.

jeffbaumes:
LGTM. I'm adding an issue to avoid the console error when navigating away from the Resize Table example but that can be resolved in another PR.

PR #100, opened 2016-04-11
Initial version of UpSet

jeffbaumes:
Depends on #97. Look at final commit to see diff.


ronichoudhury:
Congrats on PR #100! :tada: :balloon: :notes:

src/candela/components/UpSet/index.js
```diff
@@ -0,0 +1,90 @@
...
+    let data = [];
+    let header = [this.options.id];
+    this.options.sets.forEach(s => header.push(s));
+    data.push(header);
+    this.options.data.forEach(d => {
+      data.push(header.map(h => '' + d[h]));
+    });
+    console.log(header);
+    console.log(data);
+    console.log(this.options);
```
ronichoudhury: Do these ``console.log()`` statements need to stay?

package.json
```diff
@@ -51,8 +51,11 @@
     "karma-webpack": "^1.7.0",
     "legacy-loader": "0.0.2",
     "less": "^2.3.1",
+    "node-sass": "^3.4.2",
```
ronichoudhury: Is this actually used somewhere? I figured ``sass-loader`` would have taken care of that dependency?

ronichoudhury:
I noticed that the styling on the examples listing has changed - is that because UpSet is somewhat indiscriminate with its CSS (i.e., is that due to #101)?

(For the record, the example listing looks nicer now than before :tophat:)

jeffbaumes:
Yeah we need to fix #101 - that is what is going on, but since it looks ok I was going to resolve that separately from this PR.

package.json
jeffbaumes: `npm i --save-dev sass-loader` output a warning requiring a peer dependency of node-sass at 3.4.2, so I obeyed npm's output and added it. Not that I fully understand the nuance of peer dependencies.

src/candela/components/UpSet/index.js
jeffbaumes: These should go away. I'll push a commit.

jeffbaumes:
Ok debug statements removed. Ready for another review.

ronichoudhury:
I had a line note above about ``node-loader`` and ``sass-loader``... aside from that concern, this LGTM.

jeffbaumes:
@ronichoudhury I replied to your line note - does that logic seem ok to you?

package.json
ronichoudhury: I thought peer dependencies might be involved :smile_cat:

I basically just wanted to know there was a reason for including it. Thanks!

ronichoudhury:
Logic looks good. LGTM.

PR #107, opened 2016-04-12
Initial Jupyter support

### LINKED ISSUE #94: Jupyter bindings
So this happened today after a few minutes of playing around

screen shot 2016-04-06 at 7 35 37 am

I think it would be quite easy to build a tiny Python library that did this without needing JS code. The bindings would have easy ways to allow pandas dataframe input and we could easily pass a Python dict of options through to JS.

jeffbaumes:
Fixes #94


ronichoudhury:
I love it. LGTM.

PR #109, opened 2016-04-13
Moving to strict padding

### LINKED ISSUE #85: Switch to "strict" padding for Vega components
This is a newer Vega feature that will auto-compute the padding and keep the overall width and height intact. This will eliminate the need for special padding properties and logic on Vega visualizations.

jeffbaumes:
Fixes #85. This solves several things:

 - Simplifies all Vega specs to eliminate special padding logic

 - Prompted the use of the new width and height signals so all pan/zoom axes are simpler now

 - Biggest one: we were rendering Vega specs twice. Once to get the padding
   to use for fixing up the width and height, then again for the final render.
   With strict mode this is no longer needed, removing some hacky code.

### DIFF
--- a/src/candela/components/Bar/spec.json
+++ b/src/candela/components/Bar/spec.json
@@ -10,7 +10,7 @@
   {
     "width": ["@get", "width", 800],
     "height": ["@get", "height", 500],
-    "padding": ["@get", "padding", {"top": 20, "bottom": 200, "left": 100, "right": ["@if", ["@get", "color"], 150, 10]}],
+    "padding": "strict",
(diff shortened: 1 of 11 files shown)


ronichoudhury:
I assume this PR doesn't affect the look of the current examples, just internal logic, right?

If that's the case, then LGTM.

jeffbaumes:
That is true. The only change is that things are tighter now in general since there is no guessing or overestimation about padding size to make sure things fit. I'm relatively confident there are no regressions - I clicked through each of the examples and they looked good.

PR #110, opened 2016-04-13
Adding some logic to guard against leaving the resize-table example

### LINKED ISSUE #102: Console error when navigating away from the Resize Table example
We should cancel the repeating timer by using `setTimeout` instead, or at least guard the code with an `if` so that the console error is not generated.

jeffbaumes:
Fixes #102.


ronichoudhury:
LGTM.

PR #114, opened 2016-04-14
Update UpSet to version with scoped css

### LINKED ISSUE #101: Update UpSet CSS not to pollute global style
Currently UpSet defines these (and probably more), they should be made to only be used within UpSet element.

```css
a:link, a:visited {
    color: black;
}
* {
    margin: 0;
    padding: 0;
}
```

jeffbaumes:
Fixes #101.

### DIFF
--- a/package.json
+++ b/package.json
@@ -76,7 +76,7 @@
     "jquery": "^2.2.0",
     "LineUpJS": "git+https://github.com/Caleydo/lineup.js#51ae6f2bcb55aead3f63357864e90d5799789910",
     "underscore": "^1.8.3",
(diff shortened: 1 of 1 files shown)


ronichoudhury:
This seems to work properly, but should we independently reapply the style that made the examples listing look better?

jeffbaumes:
Perhaps, but I feel that's a completely separate issue. It was messing with table striping and who knows what else that we would not want to reproduce. I think what's listed in #101 would make it look like it did before, although knowing what was visited may be worth keeping - new examples light up.

ronichoudhury:
Ok. Let's leave it as is for now, then. We may want to just style it independently in any case, which we could do in a separate (low priority) issue.

This PR LGTM.

PR #116, opened 2016-04-14
Don't npmignore the src dir

jeffbaumes:
This enables chained webpack import/require of sections of the candela repo.

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -1,3 +1,2 @@
 app
 build
-webpack.config.js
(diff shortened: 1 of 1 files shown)


ronichoudhury:
LGTM, but I added a commit to clean up a stray directive in the file from before.

This will also expose the config directory, but I think that's probably fine, since someone may need to look at out webpack configs to properly link in Candela with their own project.

PR #117, opened 2016-04-14
Add fields option to input sets from any categorical field

### LINKED ISSUE #115: Let UpSet view work with non-boolean fields
Instead of assuming 0/1 fields, generate sets from any field by creating a series of sets named "field_name value" for each value found in the field.

jeffbaumes:
Fixes #115

### DIFF
--- a/app/examples/visualizations.json
+++ b/app/examples/visualizations.json
@@ -194,7 +194,8 @@
         "Blue Hair",
         "Duff Fan",
         "School"
-      ]
+      ],
+      "fields": ["School"]
     }
   },
   {
--- a/src/candela/components/UpSet/README.md
+++ b/src/candela/components/UpSet/README.md
@@ -2,6 +2,41 @@
 
 An [UpSet](http://www.caleydo.org/tools/upset/) set visualization.
 
+UpSet interprets binary columns (i.e. columns with a literal `"0"` or `"1"`
+in every row) as sets. Any field in the `sets` option will be interpreted in
+this way. Since most data is not arranged in binary columns, the visualization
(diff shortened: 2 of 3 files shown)


ronichoudhury:
Can you explain exactly what this does? I am not too clear on it even after reading the diff and the issue it solves.

jeffbaumes:
Certainly. UpSet only works on binary columns, i.e. columns with a literal "0" or "1" in every row indicating set membership. Any field in the `sets` option will be interpreted this way. Since most data is not arranged like that, I enable preprocessing of arbitrary categorical fields with the `fields` option. Each field specified in this list will first be preprocessed into a collection of 0/1 columns that are passed to UpSet.

For example, suppose the data table is:

| id | f1 | f2 |
| :-- | :-- | :-- |
| n1 | 1 | x |
| n2 | 0 | x |
| n3 | 0 | y |

You could create an UpSet visualization with the following options:

```js
new UpSet({
  data: data,
  id: 'id',
  sets: ['f1'],
  fields: ['f2']
});
```

This would preprocess the `f2` field into `f2 x` and `f2 y` sets as follows and make them available to UpSet:

| id | f1 | f2 x | f2 y |
| :-- | :-- | :-- | :-- |
| n1 | 1 | 1 | 0 |
| n2 | 0 | 1 | 0 |
| n3 | 0 | 0 | 1 |

You can also play with it in the reference app to see how these things work. I'll put docs like this into the UpSet component docs.

jeffbaumes:
I updated the docs, ready for re-review.

ronichoudhury:
Ah ok, makes total sense :-D.

LGTM!

PR #118, opened 2016-04-15
Release 0.3

ronichoudhury:
Basing this PR against the new ``release`` branch, which will only contain commits for tagged releases.


ronichoudhury:
@jeffbaumes the only other reference to a version ``0.2.0`` in the source code is in the ``pycandela`` directory. Should we update that at this time to sync with this release?

jeffbaumes:
Up to you - I'm ok with that just sitting there for now unless you plan to also host it on github.io which in that case it makes some sense to update the link. I guess pushing a hosted bundle somewhere should be part of the release process eventually but it does not need to be now.

jeffbaumes:
LGTM. I am assuming "bump version number for release" is the only actual new commit here and that seems fine.

ronichoudhury:
The other necessary commit will be to build the dist files. That commit
will be merged to the new release branch, while the bump version commit
will be merged back to master (simply to avoid merging the built libraries
back to master). A tiny bit messy but not too bad.

jeffbaumes:
As long as the process is documented, it can be as convoluted as you like :) There is always weirdness when you have files (like the bundle) that is checked in "sometimes".

ronichoudhury:
Right. My next task on master will be to document this actually :-D

ronichoudhury:
@jeffbaumes PTOML

jeffbaumes:
ITOML, and LGTM! I verified that I can include and use the bundle in a basic HTML page.

PR #121, opened 2016-04-19
User redesign

alex-r-bigelow:
Adds:
- much more robust user log in / log out + toolchain management
- much more precise / robust / efficient routing, event triggering


jeffbaumes:
Clicking on the 0/0 text on the Mapping view gives the console error:

```
Uncaught TypeError: _this.panel.toggle is not a function
```

jeffbaumes:
There is no way to click away from the top-right menu after opening it. It seems you need to select an option in the menu before doing anything else.

jeffbaumes:
After editing a toolchain, going into its settings and clicking "New Empty Toolchain" makes a new toolchain but does not close the dialog.

jeffbaumes:
Pasting a permalink in the title bar and pressing enter does not close the currently opened dialog. E.g. if the welcome panel (3 choices of start with data, etc.) is open, pasting the permalink opens the toolchain in the background but the window remains open and there is no way to close it.

alex-r-bigelow:
The first three should be fixed. Working on the routing issue...

alex-r-bigelow:
... and the pasted URL issue should be solved now.

jeffbaumes:
When you aren't logged in and in an empty toolchain, when you click on the "add data" button to open the data dialog, there is the following uncaught error on the console:

```
Uncaught (in promise):
Object {readyState: 4, responseText: "{"message": "You must be logged in.", "type": "access"}",
responseJSON: Object, status: 401, statusText: "Unauthorized"}
```

Can we avoid this error?

jeffbaumes:
There appears to be some state-clearing issues on the mapping UI. Here are the steps to show the issue:

1. New empty toolchain
2. Open mapping panel
3. Add data - select iris
4. Add visualization - select scatter
5. Link up the top two columns with x and y
6. Go to toolchain properties
7. New empty toolchain - mapping panel should be open then eventually clear
8. Add data - select iris
9. Add visualization - select scatter
10. Start linking up the top two columns with x and y. You should see the blue dotted lines not clear after the second click to pair to a visual attribute.

I have also seen a field appear highlighted green right after selecting the dataset after previously opening that dataset. Makes me think the D3 or other state needs to be more fully cleared before reinitializing the mapping view with new data/vis.

alex-r-bigelow:
That behavior should be fixed now. And thanks for pointing it out—it also exposed a much bigger problem behind the scenes with making too many anonymous copies of library datasets (also fixed now).

alex-r-bigelow:
... and I just discovered some routing problems, so don't bother looking at this until I fix them.

alex-r-bigelow:
Okay, I *think* it's PR-worthy again?

ronichoudhury:
I'm on the user-guide branch, and I'm getting this error message when I click on the "info" icon:

````
index.js:149 Uncaught TypeError: window.mainPage.currentUser.preferences.observeTips is not a function
````

alex-r-bigelow:
If we're already discussing the user-guide branch here, I'll just go ahead and merge it.

ronichoudhury:
:+1: If this PR is meant to incorporate those changes, then I think it's
good to do that now rather than later.

alex-r-bigelow:
@ronichoudhury I can't seem to replicate that error. Any clues? (I assume it's the main info icon, not one of the ones on the widgets?)

It's also possible your user's rra_preferences item was corrupted between commits... if you trash it, the error might just go away.

ronichoudhury:
I believe Alex is still working on one final feature for this PR (help card placement in the "info mode") but in the meantime I'll just mention that I played around with this branch and everything seems to be working pretty well. It's hard to code review such a large PR, so in place of that I'm offering a tentative "LGTM" based on the fact that it's behaving well and that Alex knows better than I do what the status of the various code changes is.

alex-r-bigelow:
Yes, aside from the info mode stuff, this version should have fewer hiccups and more features than the version of the app that's currently on master.

It's not perfect, but bugs should be rare enough that we can fix them with their own branches.

ronichoudhury:
If it's ready now, I'd say merge this PR and open a new one for the Cola stuff. @alex-r-bigelow what do you think?

alex-r-bigelow:
That would be fine with me if we're cool with all the info stuff being messy.

ronichoudhury:
(Just chatted with Alex IRL; decided to wait until the Cola stuff is ready before merging.)

alex-r-bigelow:
@ronichoudhury If it LGTY, it LGTM. I've started a separate branch for actual tutorials instead of just having help bubbles, but that can have its own PR.

ronichoudhury:
I poked around and this looks nice.

I ran into a few issues which aren't enough to delay merging this PR; I'll file separate issues for each of those.

The only thing I might change before merging this is to display the name of the app as "Resonant Laboratory" rather than "Resonant Laboratory App". This would need to be changed in the splash screen, the HTML title, and the name of the Girder collection backing the app. If you git grep for 'App' you might find other instances as well.

Aside from that, LGTM!

PR #132, opened 2016-04-19
Remove stale demo outputs from webpack configuration

ronichoudhury:


jeffbaumes:
Yep, those are not needed. LGTM.

PR #136, opened 2016-04-22
Enable Travis CI

ronichoudhury:
I now have Travis running on push, performing both unit tests and image tests. I've raised the image test thresholds high enough to allow all tests to pass until we can figure out how to get more consistent images between our various development machines (I will open an issue to that effect).

There is an issue where all the unit tests are not running but I will investigate that in a separate topic.


jeffbaumes:
Other than fixing the image tests (which can be on this or another PR) LGTM.

ronichoudhury:
I'll fix the image tests on a different branch.

PR #137, opened 2016-04-23
Promises and other cleanup

jeffbaumes:
A set of fixes / features:

* `tests.bundle.js` was not correctly excluding the image tests during unit testing, which caused a Phantom error and not all tests were run. This is now fixed.
* VisComponent now stores the second constructor argument in `this.options` for convenience for subclasses.
* The Vega util is no longer an old-style pseudoclass with a state, since our components now can track state. Instead it is a single `parseChart` function that returns the Vega chart as a promise after translating the spec template and rendering it with Vega.
* Now all the Vega-based charts have the correct superclass (yay)
* Fixed up the Vega unit tests to import as `vega` instead of `vcharts`. Now hardly any vestiges of `vcharts` in the codebase.


jeffbaumes:
@ronichoudhury I directly used promises while I see you include Promise from bluebird elsewhere. Should my method be ok or should I use bluebird?

ronichoudhury:
I think the ES6 runtime provides a `Promise` class; the bluebird one is specifically to allow for changing some unfortunate runtime behavior of the stock `Promise`, so your direct usage should be ok, afaik.

src/candela/util/vega/index.js
```diff
@@ -383,60 +383,45 @@ let merge = function (defaults, options) {
...
+    el: el,
+    renderer: curOptions.renderer
+  };
 
-    that.spec = transform(that.template, curOptions);
+  let vegaSpec = transform(spec, curOptions);
 
-    vg.parse.spec(that.spec, function (chartObj) {
+  // Return a promise for the Vega chart object
+  return new Promise((resolve, reject) => {
```
ronichoudhury: Since this promise never calls `reject()`, can we just omit the `reject` argument? Maybe not; the other form for `Promise()` is to take an "executor" function, and I'm not sure how the Promise object can tell the difference between the two types of argument.

ronichoudhury:
I think this looks pretty good; I just have a few observations/opinions:
- I would have preferred that `options` stay out of `VisComponent`, since I really was striving to keep that class as simple as possible, deferring any excess complexity (including option-saving, etc.) to subclasses. If you think it's a sufficiently "core" thing to track options, I can be convinced otherwise.
- We shouldn't worry about it for this PR but, generally speaking, I'd prefer if PRs for different topics remained separate, even if a sequence of PRs ends up each depending on the last, because it makes review and post-facto searching easier (in this case, I think the fix for the testing infrastructure (nice catch), the `this.option` tracking, and the `Promise`-based update to the Vega utility module would all be their own PR).

src/candela/util/vega/index.js
alex-r-bigelow: I think this *is* using the executor function form — the `(resolve, reject) => {}` is the executor function. I think you'd be just fine to drop the `reject` argument.

The more common form for promises is usually something along the lines of
```
Promise.resolve( ... some async "thenable" object like an ajax call ...)
.then(...)
.catch(...)
```
but this doesn't work for vega-style callback arguments

ronichoudhury: You're right; I misinterpreted the (IMO, badly formatted) section header in the MDN docs.

In light of that, I'd say drop the `reject` argument, unless keeping it aids readability or comprehension or something (it probably doesn't).

ronichoudhury:
@jeffbaumes I added a couple of commits; PTAL.

ronichoudhury:
Rebased against #140 to get meaningful Travis CI results.

jeffbaumes:
LGTM

PR #140, opened 2016-04-26
Make all tests pass on Travis

ronichoudhury:
This PR conditions the image testing to be more reproducible across computers. The main changes are:
- all image tests for Vega charts are done using the SVG renderer, so that the fonts used can be controlled
- all such fonts are are set to Open Sans (from Google Web Fonts)
- all Vega charts are now padded `strict` by default, but may take a `padding` parameter when constructed
- all image tests of Vega charts now have carefully selected `padding` arguments that keep the visualizations at a size of 800x600 in a reproducible way (the tests were failing previously because `padding: "strict"` seems to behave very differently from instance to instance)

As a result, we can now set the image testing threshold to 0.1% on both my development machine and Travis, with all tests passing.

This PR also includes a one-character fix that was preventing the unit tests from executing correctly. As a result all tests now pass on Travis.


ronichoudhury:
Woohoo! @jeffbaumes check it out!

jeffbaumes:
...... jaw drops in amazement .....

LGTM! I have an issue to file about image scaling (image tests not passing for me) but certainly this is a great start.

PR #142, opened 2016-04-27
Add a structure test for built candela.js file

ronichoudhury:
Depends on #140


jeffbaumes:
Awesome. LGTM. +1 that there are now exactly 100 unit tests (dare we make more?)

PR #143, opened 2016-04-28
Use karma quiet reporter

ronichoudhury:
I thought using this reporter might fix some of the buffering-type issues we observe when running the unit tests. It doesn't, but this is still a simpler approach to reporting than what we had before, and additionally the PR cleans up some other Karma stuff we weren't using.


ronichoudhury:
Depends on #140

jeffbaumes:
LGTM

PR #156, opened 2016-05-02
Tutorial mode

alex-r-bigelow:
Fixes minor UI bugs, adds empty state hints


ronichoudhury:
@alex-r-bigelow, now that I've merged in my extant PRs, do you mind rebasing this branch against master so that the Travis tests will pass?

ronichoudhury:
- I'm not too sure about the message that reads `These projects are hidden to everyone, unless you have specifically given them permission in the Girder interface`. Something simpler like "manage permissions" with a link to the Girder folder might be better (a la KISS principle)
- The message that says `Your project is over here. Click a button or one of the bars to get started.` feels disorienting (despite the fact that I happen to already know how the application works), since "here" generally has an ambiguous referent in this case (somewhat-but-not-really related article about ["click here"](https://www.smashingmagazine.com/2012/06/links-should-never-say-click-here/))
- As much as I love the Arrested Development references, I think we should keep them subtle if they're going to be there at all, which unfortunately the Loose Seal isn't :crying_cat_face:
- For some reason, the help layer takes a fairly long while to show up when I click the question mark button. I assume this is Cola's fault but I'm not sure. Can you experiment with reducing the number of iterations Cola uses to initialize itself? (Also, can you remind me what the magic console command is to show the tuner?)
- The help layer seems a bit confusing still. For example, I see a card pointing to the visualization selector button that says "Step 2" at the beginning but there's no "Step 1" card. More generally, I'm starting to think that a "guided tour" mode (Google products make heavy use of this when, e.g., a new feature or app is introduced) might be better than the help layer, especially if Cola isn't making it easy to get the optimal layout for the help cards. The idea would be to use a modal layer to darken most of the screen except for one element, accompanied by descriptive text (these could be "Step 1" types of things). The user could click "next" to continue the tour, or "got it" to break out. I realize this one is a bigger request so we should talk about it at tomorrow's meeting.

Despite my long comment, I want to mention that this is looking better and better! Nice work so far :smile:

jeffbaumes:
When logged in, my `UserPreferences` `'seenTips'` metadata is `undefined` (probably because I was logged in as a user from a prior version that did not have that metadata). This causes a console error when trying to access it. Seems there should be more guards against that meta field being available.

jeffbaumes:
I get the following when trying to save a copy of Bar Chart

screen shot 2016-05-03 at 9 48 02 am
 
Related problem: I cannot seem to select or copy text in the modal that shows up (hence the need for a screenshot here).

jeffbaumes:
The help and hamburger icons seem a bit too large - would be nice to have more breathing room like the rest of the toolbar (note the lack of space at the top). The Dataset icon in the toolbar could probably stand to be smaller too.

screen shot 2016-05-03 at 9 55 14 am

ronichoudhury:
+1 to unselectable error message text. I ran into that yesterday but forgot to mention it.

alex-r-bigelow:
@jeffbaumes , the `seenTips` error seems odd. Even if it's missing in Girder, the Backbone model *should* be initializing it with an empty object, and things should be automagically cleaned up in Girder the next time it's saved. Any chance you have a stack trace for the error (or a way to reproduce it)?

alex-r-bigelow:
Aside from Jeff's metadata issue (added an issue that we'll resolve in a different PR), this should be ready for review.

jeffbaumes:
I agree about addressing the metadata thing separately. I assume I would not have it when starting from a clean slate, so it's more of a backwards-compatibility thing that we don't need to be concerning ourselves with too much right now.

jeffbaumes:
I think I actually fixed it by just changing the name from `rra_preferences` to `rl_preferences`. I pushed that commit. @alex-r-bigelow if that small change looks good to you, LGTM and you can merge.

ronichoudhury:
Hold off a moment before merging please... I'm still poking around :smile_cat:

ronichoudhury:
Ok, hold released :)

This looks really nice! I found some minor bugs but I will just open new issues for them - no need to restrain this PR in the meantime.

(Just to make it official: LGTM)

alex-r-bigelow:
@jeffbaumes: it makes sense that that would fix it - you could also have fixed it simply by trashing the existing `rra_preferences` item in Girder. But the problem of dealing with old/bad metadata is still there.

I should probably switch all the holdover `rra` names to `rl` throughout the app if we're going to merge that change in now...

alex-r-bigelow:
As the last change was pretty minor (just renaming events), I'm going to assume the LGTM still applies...

PR #161, opened 2016-05-02
Avoid ephemeral ports for testing

ronichoudhury:
28000 doesn't lie in the ephemeral port range on Linux, while 38000 does.

Thanks @manthey for bringing this to my attention.


manthey:
LGTM.

Issue #59, opened 2016-03-25
UI indication of multiple fields


ronichoudhury:
In the mapping interface, it would be neat if, upon selecting a field, the color or style of the lines indicated whether clicking on a property will *replace* any currently existing mapping on that property, *add* to that mapping, or some other possibility I haven't thought of.

For replace vs. add I think that dimming any existing mapping (to indicate it will go away) would pretty much satisfy this issue. Let me know what you think @alex-r-bigelow and @jeffbaumes.

PR #171, opened 2016-05-03
[WIP] Geojs component

aashish24:


config/webpack.config.js
```diff
@@ -27,6 +27,9 @@ module.exports = {
       d3: path.resolve(__dirname, 'node_modules/d3/d3.min.js')
     }
   },
+  externals: {
```
aashish24: I will take it out

aashish24:
* Get the line and polygon feature working
* Roni will make a pass on spec and option < we will discuss it some more>

ronichoudhury:
This is very exciting :)

I'll take charge of QA for this PR and will work with @aashish24 to get things ready.

aashish24:
thanks @ronichoudhury. I cleaned up the text that was not required.

ronichoudhury:
@aashish24 do you mind rebasing your branch against master, to get Travis builds and image testing working?

src/candela/components/Geo/index.js
```diff
@@ -0,0 +1,33 @@
+import $ from 'jquery';
+import geojs from 'geojs/geo.js';
+import VisComponent from '../../VisComponent';
+
+export default class Geo extends VisComponent {
+
+  constructor (el, options) {
+    super(el);
+    console.log(options);
```
ronichoudhury: Can we get rid of this line?

aashish24:
@ronichoudhury will do and also will take out the debug message.

src/candela/components/index.js
```diff
@@ -10,6 +10,7 @@ import LineUp from './LineUp';
 import ParallelCoordinates from './ParallelCoordinates';
 import Scatter from './Scatter';
 import ScatterMatrix from './ScatterMatrix';
+import Geo from './Geo';
```
jeffbaumes: Alphabetize please

src/candela/components/index.js
```diff
@@ -23,5 +24,6 @@ export default {
   LineUp,
   ParallelCoordinates,
   Scatter,
-  ScatterMatrix
+  ScatterMatrix,
+  Geo
```
jeffbaumes: Ditto

jeffbaumes:
Very excited to see this going in. We will also want some simpler modes that Resonant Lab app can pick up more easily from tabular data (e.g. a dots feature, maybe some state and country level choropleth features, etc.).

ronichoudhury:
So, this actually looks pretty complete to me for a start at getting GeoJS into the Candela family. This incarnation accepts a more monolithic feature spec that currently works for specifying any feature layer that only needs position data to work (e.g., points/dots). @aashish24 has expressed plans to make it work for all feature types. However, when linting/testing passes on Travis, I think we could more or less accept this as is and work on improvements in future PRs. I myself would use this as a springboard for exploring the more modular approach I've discussed with Aashish, which would include the "simpler modes" that Jeff mentioned above.

We should add an image test as well, but even that might come in a small PR following this one (as I'll need to tweak the settings a bit to get something reproducible for Travis).

Let me know what you think.

alex-r-bigelow:
Agreed, I'm pretty excited about this too!

I can totally see state / country categorical strings matching up nicely without any change to Resonant Lab. But for points, I'm guessing we'd want a vis option that expect exactly two numbers? Then again, there wouldn't be a good way for the vis to tell which is latitude and which is longitude - so they should probably be separate options (but may need some kind of special relationship to each other in the matching interface!)

I could see adding some of this to #59, or opening a separate issue

jeffbaumes:
@alex-r-bigelow I think @aashish24 and the Minerva team have thought a lot about finding lat and lon from tabular or other data so it's worth discussing with them. It would be simple to just make the UI match one numeric field to lat and another to lon, but there are probably better approaches. This could be a case where just a bit of smarts come in to play in detecting fields that look like locations. This could be an extension of the type sniffing that Vega's datalib is doing for us already. We also need to be thinking about supporting GeoJSON, which has explicit notions of locations. GeoJSON is a fundamental format in Minerva.

src/candela/components/Geo/index.js
```diff
@@ -0,0 +1,33 @@
...
+    this.plot = geojs.map({
+                  node: el,
+                  zoom: 6,
+                  center: {x: 28.9550, y: 41.0136}
+                });
```
jbeezley: This will actually create a memory leak because the map (by default) binds a handler to the global resize event.  You need to clean it up by calling `map.exit()` or set `autoResize=false` to prevent handling resizing automatically.

ronichoudhury:
+1 to Jeff's comments about the Minerva team's experience. If I had to do something right now about picking out lat/long values, I'd probably stick with the de facto way the other Candela components have been doing it - namely, specify `latitude` and `longitude` field names. But extending our datalib-type approach sounds promising for guessing which fields might possibly be geolocations.

src/candela/components/Geo/index.js
jeffbaumes: This raises the larger issue of removing handlers in general. We may need a `destroy` (or `exit`) method on vis components to let them handle things like this.

ronichoudhury: +1. See #174.

ronichoudhury:
Does anyone know how to get Travis to build/test Aashish's branch? Does someone have to check out Aashish's project and push a remote here? Is there a better way?

ronichoudhury:
@aashish24, could you push your branch to `origin` and open a new PR for that branch? Once that's done we can link that PR to this one to preserve the discussion thread, and then close this one.

ronichoudhury:
@aashish24 I opened #198 in favor of this PR; could you close this one when you get a chance?

aashish24:
Sure thanks @ronichoudhury closing this one.

PR #179, opened 2016-05-04
Get rid of one-off "LineChart" class

ronichoudhury:
This also recreates the "dynamic line chart" example using a custom `Events`-mixedin class extending `Line`.


jeffbaumes:
LGTM!

PR #183, opened 2016-05-05
Rename classes with more descriptive names

ronichoudhury:


jeffbaumes:
LGTM. Note that this invalidates all saved projects in Resonant Lab. But better now than later.

PR #191, opened 2016-05-06
Add codecov.io support

ronichoudhury:
Depends on #185.


jeffbaumes:
Awesome. LGTM. It works with codecov Chrome plugin also:

screen shot 2016-05-06 at 1 56 59 pm

PR #195, opened 2016-05-09
Fix crash when opening some new datasets

alex-r-bigelow:
Well here it is... my first one-line PR

### DIFF
--- a/app/resonant-laboratory/models/Dataset.js
+++ b/app/resonant-laboratory/models/Dataset.js
@@ -19,6 +19,9 @@ let VALID_EXTENSIONS = [
 ];
 
 let dictCompare = (a, b) => {
+  if (!(a instanceof Object) || !(b instanceof Object)) {
+    return false;
+  }
   let aKeys = Object.keys(a);
   let bKeys = Object.keys(b);
   if (aKeys.length !== bKeys.length) {


ronichoudhury:
FYI, you still need to wait for code review and an "LGTM" from someone else
before merging, even for small pull requests.

TristanWright:
I don't know the nature of the crash, but if you're checking that `a` or `b` are `{}` you might want to use a different check. `x instanceof Object` is truthy for a [number of other types](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/instanceof#Examples): String, RegExp, Array...  

([I use this one](https://github.com/tjmehta/101/blob/master/is-object.js))

jeffbaumes:
Thanks @TristanWright. Does HPCCloud or other projects depend on 101 as a util library? Hadn't heard of that one before.

ronichoudhury:
@TristanWright thanks for pointing that out - I was worried about something along these lines, since JavaScript is notorious for its type system absurdities. Obligatory video link (only 4 minutes long): [WAT](https://www.destroyallsoftware.com/talks/wat).

ronichoudhury:
101 looks interesting; thanks for sharing!

alex-r-bigelow:
In this case, it's pretty moot... on the odd chance a String, Array, or whatever comes in, it will still behave appropriately (it's just making sure that `Object.keys()` won't throw an error). Perhaps I should have just said `if (!a || !b)`, but that felt less descriptive.

TristanWright:
haha yes WAT!

@jeffbaumes we use a few things in [mout](https://github.com/mout/mout/tree/master/src) ([only a few](https://github.com/Kitware/HPCCloud/search?utf8=✓&q=mout)). 101 is a bit more functionally oriented. Whenever I do need something from 101 it's so small I end up copying just the function into the code base and commenting its source. It's often `is-object` too but I've used `pluck` and `pick` once or twice.

Issue #165, opened 2016-05-03
Better scrolling behavior in matching interface


alex-r-bigelow:
Collapse data types, possibly arrange data attribute side as grid

alex-r-bigelow:
This was fixed in 831e305b10c5f0becf2aa447a35373c63cfc7996

PR #196, opened 2016-05-09
Matching redesign

### LINKED ISSUE #59: UI indication of multiple fields

alex-r-bigelow:
This is a redesign of the matching interface. It's not everything we want yet, but should be enough for user testing (we should probably wait for multiple dataset support and/or breaking the matching interface out as an independent component before doing a deeper redesign).
Partially addresses #59 and #165


app/resonant-laboratory/views/widgets/MatchingView/index.js
```diff
@@ -5,6 +5,22 @@ import myTemplate from './template.html';
...
 import './style.css';
 
+import booleanIcon from '../../../images/boolean.svg';
+import integerIcon from '../../../images/integer.svg';
+import numberIcon from '../../../images/number.svg';
+import dateIcon from '../../../images/date.svg';
+import stringIcon from '../../../images/string.svg';
+import stringListIcon from '../../../images/string_list.svg';
+
+let ICONS = {
```
ronichoudhury: This should probably be `const` instead of `let`.

app/resonant-laboratory/views/widgets/MatchingView/index.js
```diff
@@ -5,6 +5,22 @@ import myTemplate from './template.html';
...
+  boolean: booleanIcon,
+  integer: integerIcon,
+  number: numberIcon,
+  date: dateIcon,
+  string: stringIcon,
+  string_list: stringListIcon
+};
+
 let NODE_MODES = {
```
ronichoudhury: s/`let`/`const`/

ronichoudhury:
This looks/feels awesome! I just have two style suggestions:

1. The visual behavior for deselecting a matching is a little unintuitive. I think it should basically be the same behavior as selecting a matching, but in reverse. When making a match, you see potential links as light blue, dotted lines; hovering over a potential target turns it into a dark blue, solid line. Clicking to select turns it light green, solid, and removes the dotted lines. When *de*selecting I think the green line should turn light green and dotted, and upon clicking to deselect it should disappear (as it does currently).

2. Is there a less bulky icon concept we can use for the Boolean type? The Venn diagram icon is "inky" and it connotes sets more directly than Booleans.

Both of these could be handled in separate topic branches (if they are deemed valid suggestions), so let's discuss how we want to handle them.

alex-r-bigelow:
I totally agree with both suggestions... actually, it looks like something with the line hovers / highlights is misbehaving (more software bug, less design flaw), so I'll get on fixing that (changing to WIP until that gets fixed). And I'll try to find a better boolean icon.

jeffbaumes:
Looks like now when deselecting it turns orange and dotted to indicate the link will be removed. This seems fine to me. Boolean icon is much better - I just think it needs to be compressed more so it looks more like one icon and doesn't run into other icons as much:

screen shot 2016-05-13 at 9 04 29 am

Other than that one icon thing LGTM, so if you fix it and ping me for final validation, you should be good to merge.

alex-r-bigelow:
If this LGTY, I'll go ahead and merge:

screen shot 2016-05-13 at 9 44 40 am

jeffbaumes:
LGTM. Bonus points for extra padding all around.

PR #199, opened 2016-05-10
Tutorial mode

alex-r-bigelow:
We may still want to add/change the order of specific tips, but this should at least be production-quality now.


ronichoudhury:
This looks pretty good. Some questions/suggestions:

1. Is this compatible with what we decided yesterday about putting individual toggleable tips on the various pieces of the UI (i.e., help bubbles that are not part of this tip sequence)?
2. There is an odd pause before the bubble moves from "rename this project" to "main menu" that should be eliminated.
3. This may sound weird but I think the animation in this case is actually less helpful than just bopping the circle-and-arrow and the focus ellipse to the next tip without animation. It would make it feel more like a quick refresher on the UI, whereas the animation kind of makes it feel sluggish somehow, like we're wading through molasses to get to the next piece of UI.

ronichoudhury:
I'm also not sure if it's worth the extra code complexity to draw a distinction between what the user has and hasn't seen before. The tip sequence is fast enough that it doesn't really matter, on top of which repetition is good for learning a new system, so we could promote that by not necessarily reminding/admonishing the user that they've seen something before.

alex-r-bigelow:
This is great feedback; thanks! I'll speed up or remove the animations. The pause probably has something to do with the interface skipping tips that jQuery can't find on the screen... but you're right that it shouldn't affect the speed.

With storing the tips the user has seen, I think the main thing I cared about was the blue icon as a more subtle hint that there's new stuff to check out—the blue bubbles were extra (and I'm happy to take them out). I was planning on logging user achievements with the same mechanism, so the complexity will probably stay even if we stop storing the tips.

ronichoudhury:
In that case, don't worry about remove the blue/yellow backgrounds. It's not a big deal, just something I thought I'd raise for discussion.

As for speeding up vs. getting rid of animations, I'd suggest we try eliminating animation altogether to see how it feels. To me, it's not a matter of how fast the animation is, it's more that there's any animation at all.

alex-r-bigelow:
Some of the transitions felt really clunky without any animation (particularly the lightbox "hole" jumping from a small icon to span an entire toolbar), but I'll see what I can do to make it less obnoxious.

jeffbaumes:
This is awesome. LGTM. If you want to support rectangle "holes" in some situations like this one, feel free to fix and merge, otherwise you can just merge as is and open an issue to revisit at some point.

screen shot 2016-05-13 at 11 07 05 am

alex-r-bigelow:
Sounds good—issue #213 created.

PR #202, opened 2016-05-11
Stylus loader update

ronichoudhury:
This was needed to sidestep a weird build issue with stylus 0.52
(https://github.com/stylus/stylus/issues/2176), which was being used as a
dependency of stylus-loader 1.5.


alex-r-bigelow:
LGTM (as in I rebuilt without any problems—I don't actually use any stylus, though...)

ronichoudhury:
Thanks @alex-r-bigelow!

PR #210, opened 2016-05-12
More small UI fixes

alex-r-bigelow:
- Buttons that switch datasets and visualizations are now actual buttons
- Dataset info button previously fired the help interface, not the info dialog (fixed)
- Prevent listener corruption when a dataset replaces itself


jeffbaumes:
LGTM

jeffbaumes:
LGTM with your new changes.

Issue #204, opened 2016-05-12
Better dataset attribute inference / coercion


alex-r-bigelow:

alex-r-bigelow:
We should do more to correctly infer / update / allow the user to override dataset attributes.

Symptomatic / related issues:
#216 #203 #175 #169 #168 #162

Examples of current problems:
- datalib can't handle datasets that have a dict as the root instead of an array
- Coerce values when the user specifies a type (e.g. "$3.25" -> 3.25)
- Properly convert dates

Thoughts:
- It would be handy to be able to pull schema information from database items
- We should prioritize how competing sources for this information are adopted. We can learn "data attribute A is type X" from:
  - The user's direct coercion settings
  - Auto-inference (e.g. datalib)
  - More explicit sources (e.g. database schema)
  - Previous users' coercion settings for a dataset
- This is further complicated by the fact that data can be reshaped, potentially invalidating the user's settings

alex-r-bigelow:
Fixed in #290

Issue #175, opened 2016-05-04
Handle lat/lon matchings intelligently


alex-r-bigelow:
See this discussion: https://github.com/Kitware/candela/pull/171#issuecomment-216887014

Interestingly, it seems that we're pulling stuff from the vis side to learn about the data, at least conceptually (datalib is really a vega thing, minerva has its cool lat/lon detection stuff). Is there something bigger here that should be part of candela (where, given a dataset, a component can optionally suggest data types and/or matchings)?

ronichoudhury:
I think there's definitely an opportunity here. Maybe @mgrauer should add this topic to his fact-finding mission...?

mgrauer:
Let's talk about this next week.

ronichoudhury:
Excellent.

alex-r-bigelow:
Currently, Resonant Lab (technically) supports mapping mystery objects to visualizations that expect them. Beyond that, we're thinking that this would fall under the category of data reshaping, in which a separate tool should be used.

Issue #168, opened 2016-05-03
Filtering interface / endpoints


alex-r-bigelow:
Support interactive filtering of datasets via mongo / csv query endpoint(s)

alex-r-bigelow:
(@ronichoudhury is looking in to the mongo Item part of this problem)

alex-r-bigelow:
Fixed in #316

Issue #162, opened 2016-05-03
Interface for uploading datasets + simple management


alex-r-bigelow:
(similar to existing project management interface)

PR #215, opened 2016-05-13
Enable database items in Resonant Laboratory

ronichoudhury:
This is WIP because it doesn't quite work yet. I will need some help tracking down the problems and possibly redesigning some of the datatype handling stuff.

### DIFF
--- a/app/resonant-laboratory/models/Dataset.js
+++ b/app/resonant-laboratory/models/Dataset.js
@@ -89,20 +89,26 @@ let Dataset = MetadataItem.extend({
     if (cache && this.rawCache !== null) {
       return Promise.resolve(this.rawCache);
     } else {
-      return Promise.resolve(girder.restRequest({
-        path: 'item/' + this.getId() + '/download',
+      const cacheData = data => this.rawCache = data;
+
+      const databaseQuery = girder.restRequest({
+        path: `item/${this.getId()}/database/select`,
         type: 'GET',
-        error: null,
-        dataType: 'text'
-      })).then((data) => {
-        if (cache) {
-          this.rawCache = data;
-        }
-        return data;
-      }).catch(() => {
-        this.rawCache = null;
-        return null;
+        dataType: 'json'
       });
+
+      return databaseQuery.then(resp => {
+        return JSON.stringify(resp.data);
+      }, () => {
+        console.log('hello');
+        return girder.restRequest({
+          path: 'item/' + this.getId() + '/download',
+          type: 'GET',
+          error: null,
+          dataType: 'text'
+        });
+      })
+      .then(cacheData, () => this.rawCache = null);
     }
   },
   getSpec: function () {
--- a/app/resonant-laboratory/models/Dataset.js
+++ b/app/resonant-laboratory/models/Dataset.js
@@ -100,7 +100,6 @@ let Dataset = MetadataItem.extend({
       return databaseQuery.then(resp => {
         return JSON.stringify(resp.data);
       }, () => {
-        console.log('hello');
         return girder.restRequest({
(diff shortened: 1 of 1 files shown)


alex-r-bigelow:
After poking around with this a little, it looks like the problem is stemming from only using datalib to infer the attributes in the dataset (if it chokes, for now I stupidly assume that we can't figure out what the attributes are, hence the error that the interface is showing... BTW, if you merge the ui-fixes branch onto this one, the error message should work properly).

I was able to get this to work with some of the datasets we've already been working with, e.g.:
```
mongoimport --db test --collection gapminder --drop --file gapminder.json --jsonArray
```
Then, of course, create an item, and then give it its connection information via `PUT /item/{id}/database`:
```
{"url":"localhost:27017","database":"test","collection":"gapminder","type":"mongo"}
```
(you probably know most of this... this is more documentation for my benefit :-) )

To make this work in general, I think we just need to infer the attributes of the response in a more intelligent way... Actually, we should probably retrieve the *schema* of the dataset at some point anyway so that we can build the filter interface...

ronichoudhury:
The gapminder data seems to work. Let me know what happens when you try out points.json using the same procedure.

alex-r-bigelow:
I just got it to work... It may have something to do with my dumb reliance on the item name having an "extension" that reflects its format (I'll file an issue). If you rename the item to be "points.json" it should work.

ronichoudhury:
Well I did name mine `points.json`... but in my Girder view of that item, I see it lists `{}` as its `attributes` metadata entry. Does that have something to do with it?

alex-r-bigelow:
That's very possible. You could try trashing its metadata and see if it helps

ronichoudhury:
Welp. That worked beautifully. I wonder what went wrong on Friday...

alex-r-bigelow:
A lot of these problems seem to be related to this thread:
#216 #203 #175 #169 #168 #162

I've generalized #204 to try and describe the higher-level issue

app/resonant-laboratory/models/Dataset.js
```diff
@@ -89,20 +89,26 @@ let Dataset = MetadataItem.extend({
...
-      }).catch(() => {
-        this.rawCache = null;
-        return null;
+        dataType: 'json'
       });
+
+      return databaseQuery.then(resp => {
+        return JSON.stringify(resp.data);
+      }, () => {
+        console.log('hello');
```
jeffbaumes: Remove debug statement.

ronichoudhury: Oops! Removed.

alex-r-bigelow:
LGTM—for now, I think it's fine to probe database status, and let the 400 error show up on the console. I'll probably add a more direct check whether an item is a database vs a file as part of my sprint tomorrow

ronichoudhury:
In the future the hiccup about "database items" should theoretically go away. It's something I need to discuss with Zach but eventually it will all just be abstract Girder items - no special probing logic necessary.

alex-r-bigelow:
Are you planning on making database/not-database status more of a first-class attribute of an item (is it already)? I was planning on adding a flag to the item's metadata, but there's no sense in being redundant...

ronichoudhury:
No, we shouldn't make explicit whether an item is database-backed or not; it should be an abstract feature of an item. Currently the only way to autodetect whether an item has database backing is probing the endpoint and watching for an error. If putting a flag on the item makes this logic easier/cleaner, then we should do that. But in any case both approaches reflect a flaw in the current architecture, which can be solved in the long term by introducing the correct database-backing semantics to Girder items.

If this doesn't make sense let's talk IRL :smile_cat:

PR #220, opened 2016-05-16
Exclude dist and build directories from linting.

manthey:
Otherwise, you can't npm run lint in the same directory where you build candela without getting errors.

Add cobertura coverage output to work with other coverage tools (eventually we can send this to cdash).  I have a local tool which works with these files.


ronichoudhury:
LGTMITIH

PR #221, opened 2016-05-17
Fix #207 - hover point fills as well as stroke

### LINKED ISSUE #207: Scatterplot: hover point fills as well as stroke
Hovering over the center of a point should still trigger the tooltip, not just the stroke

jeffbaumes:
Fixes #207


jeffbaumes:
@ronichoudhury I think the test changed slightly (there is a slight fill now with 0.01 opacity), can you take a look? I don't know how to find the failed test image. Unless it looks bad I'd say we should check in a new image version on this branch for the scatterplot test.

ronichoudhury:
I updated the testing image, so LGTMITIH.

I'll also open an issue for documenting the testing process, including how to find the testing images for comparison, and how to check in a new baseline image.

PR #222, opened 2016-05-17
Fixes #180 - Heatmap labels change on render

### LINKED ISSUE #180: Heatmap labels change on render
The heatmap has a weird behavior - every time it renders, the labels on the y-axis can change. To reproduce:

1. Build the project (e.g. `npm run build`).
2. Start the testing server with `npm run demo`.
3. Visit http://localhost:3000/#heatmap in a browser.
4. Observe the lowest number on the y-axis (should be `429`).
5. Change the size of your browser window to force resize-based re-render.
6. Observe the lowest number on the y-axis again (should be larger than `429`).

jeffbaumes:
Fixes #180. If using the default id field, don't display the value on the y-axis.


jeffbaumes:
@ronichoudhury looks like we need an image change here too.

ronichoudhury:
I've updated the testing image, so LGTMITIH.

PR #223, opened 2016-05-17
Fully delete unused options instead of setting to null

jeffbaumes:
This bug manifests when you set then unset a property (in particular I saw this in the `id` option of Heatmap). Setting an option to `null` and removing it entirely could have different effects on the visualization logic. In this case setting the id option to `null` set all ids to `undefined`, creating a single row in the heatmap. On refreshing the ResLab page, the option was removed and the vis functioned normally again.


alex-r-bigelow:
LGTM, thanks for the fix!

PR #231, opened 2016-05-17
Update to the latest version of LineUpJS.

manthey:
The latest version of LineUpJS uses browserify, and cannot be used as distributed with webpack due to conflicts about which packager controls the require function.  There may be a way to include the dist/LineUpJS.js file directly, but I couldn't figure it out.  Rather, I've added a script which uses LineUp's grunt file to compile the typescript, and then include the compiled base file from the compiled typescript in our component.  From here, webpack can properly handle the dependencies.

We no longer use the demo css.


manthey:
@ronichoudhury If you can think of a cleaner way of including LineUp, I'd love to do it.  This uses a helper script.

Also, I'd be happy to add some unit tests and code coverage of the LineUp component, either as part of this PR or as a separate one.

manthey:
We could add additional options for LineUp.  In another application, I found it useful to be able to control which tools are available on columns (rename, filter, snapshot, hide, etc.).

jeffbaumes:
@manthey I pushed a commit to make this work with reslab. It apparently needs the stylus rule now.

jeffbaumes:
This is looking good. @ronichoudhury we need another updated image - perhaps you can document how to do this? I talked to @manthey and he is going to:

1. Update the LineUp build script so it doesn't repeat on each build.
2. Submit an issue on LineUp GitHub to ask for a way to use LineUp within a webpack build. If there is a solution then we can get rid of the LineUp build step altogether.

After those are done I feel this is good to merge.

jbeezley:
You might try to use [derequire](https://github.com/calvinmetcalf/derequire) on the lineup bundle to rename the `require` statements.  I think that is the standard solution for reusing browserified code.

src/candela/components/LineUp/index.js
```diff
@@ -83,35 +101,41 @@ export default class LineUp {
...
   createLineupAdjustWidth (elem, name, spec, fixed) {
-    let total = 0;
-    let count = 0;
-    let c1;
-    let c2;
-    // let width = $(elem)[0].getBoundingClientRect().width - $.scrollbarWidth() - fixed;
-    let width = $(elem)[0].getBoundingClientRect().width - 30 - fixed;
-    let col = spec.dataspec.layout.primary;
-
+    var rankWidth = 0;
```
ronichoudhury: Minor style thing but I'd prefer the use of `let` everywhere rather than `var` unless there's a semantics reason to prefer `var` here.

src/candela/components/LineUp/index.js
```diff
@@ -83,35 +101,41 @@ export default class LineUp {
...
+          continue;
+        }
         count += 1;
         total += this.lineupGetColumnWidth(name, col[c1], fixed);
       }
     }
-
-    let avail = width - count * fixed;
-    let scale = avail / total;
+    var avail = width - count * fixed - rankWidth;
```
ronichoudhury: Same `let`/`var` comment as above.

ronichoudhury:
So this seems to be passing on Travis now, which means everything looks good I think. @manthey can you confirm that my changes are ok?

Also, did you open a LineUp issue as described above? If so could you tag me in that issue / otherwise loop me in somehow?

Thanks!

manthey:
I filed an issue with LineUp here: https://github.com/Caleydo/lineup.js/issues/109 .

@ronichoudhury Your changes LGTM.

We should probably revisit this in a few months, as LineUp is under active development.

ronichoudhury:
Thanks for linking me to the issue.

I agree about revisiting in a few months. I'm leaving myself a reminder to do so.

PR #237, opened 2016-05-18
Initial OnSet visualization

### LINKED ISSUE #224: Incorporate OnSet component

jeffbaumes:
Fixes #224. This initial version incorporates the OnSet code directly (original is at https://github.com/tmajor86/setvis). There were several changes needed to make it modular. Since this is over 5K lines of code, we may want to think about how this is maintained going forward. The OnSet changes could be upstreamed or maintained in an OnSet fork. For simplicity this branch is a decent way to try out the integration before we decide what to do.

### DIFF
--- a/app/examples/index.json
+++ b/app/examples/index.json
@@ -99,5 +99,9 @@
   {
     "title": "Survival Plot",
     "link": "survival"
+  },
+  {
+    "title": "OnSet",
+    "link": "onset"
   }
 ]
--- a/app/examples/onset/index.js
+++ b/app/examples/onset/index.js
@@ -1,0 +1,17 @@
+import { simpsons } from '../util/datasets';
+import showComponent from '../util/showComponent';
+
+window.onload = () => {
+  showComponent('OnSet', 'div', {
+    data: simpsons,
+    id: 'Name',
+    sets: [
+      'Male',
+      'Power Plant',
+      'Evil',
+      'Blue Hair',
+      'Duff Fan',
+      'School'
+    ]
+  });
+};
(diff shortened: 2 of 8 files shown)


jeffbaumes:
I made a few more changes to isolate OnSet a bit more in the external directory, plus made it closer to the original code by doing imports differently.

ronichoudhury:
This is really cool. Can you explain what changes you had to make to OnSet? I'm not sure if you made those changes in a particular commit, or just as you moved the files into the project. Depending on what you had to do, a fork of OnSet with mainline-compatible PRs going back to the original sounds like it might be the right approach.

jeffbaumes:
Agree. I currently have an email out to the original devs to see if this is possible. I suggest we just let this branch just hang out here for a few more days.

ronichoudhury:
Roger wilco.

curtislisle:
OnSet! ....

Nice! ....

:-)

jeffbaumes:
I just force-pushed a change that references a componentized OnSet library at https://github.com/Kitware/setvis instead of relying on a copy of the OnSet code in Candela. This diff should be much more consumable now and this PR is ready for review.

@ronichoudhury please take a look when you have some time. I am pretty sure I added the OnSet example correctly but let me know if there is something off. Also, I am not sure if I should do an image test for OnSet.

alex-r-bigelow:
Very cool! LGTM as far as I'm concerned.

ronichoudhury:
This is awesome. LGTM here too.

Adding a test would be great but
- it's hard to do for this case because the default state doesn't really show anything, and
- I'd argue that testing ought to lie with `setvis` in this case anyway

If these arguments are weak (another part of me thinks they are) we can always add a test in a later PR.

jeffbaumes:
Thanks for the review @ronichoudhury.

PR #239, opened 2016-05-18
The example csvs were not being parsed as expected.

manthey:
Although vega datalib says that parse: 'auto' is the default, I found that this needs to be specified explicitly or it is not done.  For example, loading a csv file, all of the fields are strings and datalib.type.all(data) will return string for all columns.

### DIFF
--- a/app/examples/index.js
+++ b/app/examples/index.js
@@ -12,9 +12,9 @@ import visualizations from './visualizations.json';
 
 let datasets = {
   iris,
-  stocks: dl.read(stocks, {type: 'csv'}),
-  msft: dl.read(msft, {type: 'csv'}),
-  simpsons: dl.read(simpsons, {type: 'csv'})
+  stocks: dl.read(stocks, {type: 'csv', parse: 'auto'}),
+  msft: dl.read(msft, {type: 'csv', parse: 'auto'}),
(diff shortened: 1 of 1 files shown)


ronichoudhury:
Can you describe a before/after behavior so I can see the difference? Is there a way and/or need to test this behavior?

manthey:
If you examine one of the data sets (for instance, the simpsons.csv set), without this change, a loaded row looks like {"Name": "Lisa", "school": "1", "Blue Hair": "0", ... "Age": "8"}.  `datalib.type.all(data)` will return "string" as the datatype for all attributes.  With the change, the row looks like {"Name": "Lisa", "school": 1, "Blue Hair": 0, ... "Age": 8}  `datalib.type.all(data)` will return "integer" for all attributes except "Name".

I discovered this by swapping the dataset used in the LineUp example, and was surprised when all columns were treated as text.

I don't think this is affecting any of the examples, but a test wouldn't hurt.  I don't think we have any example-specific tests yet.

ronichoudhury:
I guess the test for now is whether UpSet still works properly on the simpsons data. If it does, then this LGTM.

As an aside, @alex-r-bigelow is working on a new way to do data type inference that should be more flexible. I'm not sure if datalib will still be a part of the process somewhere.

manthey:
The UpSet example works on my machine, so I'm merging.

PR #240, opened 2016-05-18
Small UI improvements

alex-r-bigelow:
This PR contains two small interface improvements that were added in the course of user studies.
- A notification layer was added (currently only shows when the user clicks the Copy Link button)
- Icons were added to the swap buttons


jeffbaumes:
Currently the swap button icon only changes color when you hover over the icon itself. I would change it to either never change color or to always change color when you are hovering over any part of the button.

Toast notification looks good!

ronichoudhury:
Very nice. LGTM!

PR #242, opened 2016-05-18
Add metadata option to UpSet

jeffbaumes:
Style is a little off in Resonant Lab for the metadata selection, but otherwise this seems to work pretty well.

### DIFF
--- a/app/examples/visualizations.json
+++ b/app/examples/visualizations.json
@@ -231,7 +231,8 @@
         "Duff Fan",
         "School"
       ],
-      "fields": ["School"]
+      "fields": ["School"],
+      "metadata": ["Age"]
     }
   },
   {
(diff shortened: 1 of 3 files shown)


jeffbaumes:
Example usage on Simpson's dataset with Age as metadata:

screen shot 2016-05-18 at 2 29 45 pm

curtislisle:
Awesome!  Thanks!  This is a great feature for supporting drill down of individual data elements in UpSet!

curtislisle:
Can we merge PR #242 on Candela soon or should it stay separate?  I'd like this in the ABCC deliverable and it increases UpSet's drill down capability.

ronichoudhury:
More or less LGTM - @jeffbaumes if you think this is ok to merge, please go ahead and do so.

@curtislisle, the other option is to include this in the VM provisioning by commit hash (unless you want to refer to `master` in the provisioning script specifically).

PR #245, opened 2016-05-19
Increase overlay z-index for overlays to be above 100

### LINKED ISSUE #228: Upset floating div
A div from UpSet floats over various Resonant Lab interface elements

jeffbaumes:
Fixes #228.

This allows visualizations (like upset) to use z-index within their visualizations (up to a value of 100 which seems reasonable).

I had trouble reproducing the error of the floating div on my machine. It looks like the floating div is meant for taking care of the fade-in of the interface, then is supposed to be transparent, but apparently it sticks around sometime. @alex-r-bigelow I know I've seen it on your machine if you want to verify it fails before this branch but works after applying this.

### DIFF
--- a/app/resonant-laboratory/views/layout/HelpLayer/style.css
+++ b/app/resonant-laboratory/views/layout/HelpLayer/style.css
@@ -1,6 +1,6 @@
 #HelpLayer svg {
   position: absolute;
-  z-index: 10;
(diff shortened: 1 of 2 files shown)


alex-r-bigelow:
Yes, this seems to have fixed it. LGTM!

PR #246, opened 2016-05-19
Changing UpSet to be smarter about finding binary sets

### LINKED ISSUE #212: Make "sets" input of UpSet smarter
Take certain strings like "yes", "true" as inclusion values.

jeffbaumes:
Fixes #212.

### DIFF
--- a/app/examples/data/simpsons-alternate.csv
+++ b/app/examples/data/simpsons-alternate.csv
@@ -1,0 +1,25 @@
+Name,School,Blue Hair,Duff Fan,Evil,Male,Power Plant,Age
+Lisa,yes,FALSE,0,0,0,0,8
+Bart,yes,FALSE,0,0,1,0,10
+Homer,no,FALSE,1,0,1,1,40
+Marge,no,TRUE,0,0,0,0,36
+Maggie,no,FALSE,0,0,0,0,1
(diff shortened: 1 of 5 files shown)


jeffbaumes:
@curtislisle you had mentioned making the binary sets smarter - do these added cases suffice or are there some other ways binary fields appear in your data?

ronichoudhury:
This looks good but can you update the documentation to explain that capitalization for the alternate terms doesn't matter?

More generally, I wonder if we can have a mode where you pass in a predicate function, and the result of running data through that function yields the 1/0 value used by UpSet?

jeffbaumes:
@ronichoudhury I changed the docs to mention that strings are case insensitive, so this is ready for re-review.

Functors returning true or false could be added at some point, but right now I think we should focus on the bare serializable API. One concern is that we will need a different testing harness if we are going to support things like function arguments, and I'd like to keep the API as simple as possible for now.

ronichoudhury:
You are right about serializability. Simplicity over generality at the current moment is the right way to go. It might make sense as an icebox issue to allow someone to specify what the binary values in their particular data set are (i.e. they could choose "si" == `true` and "non" == `false` if they are at KEU).

In any case, LGTM.

PR #249, opened 2016-05-23
Handle empty data in UpSet

### LINKED ISSUE #248: UpSet visualization sometimes doesn't come back from browser refresh
when running in Resonant Lab, UpSet always works the first time matching is established.  But sometimes UpSet just spins continually after the browser is refreshed.  It seems as though the dataflow through candela is not being refreshed.  In CSV files, assigning "boolean" variables seems to cause this. girder-db-items seem to often exhibit this behavior.

jeffbaumes:
Fixes #248. The underlying UpSet code does not seem to handle empty data well, and when loading an UpSet visualization directly from a project settings (e.g. refreshing a page) ResLab attempts to render with empty data before the data arrives asynchronously.

This fixes the issue by not rendering UpSet until data is available, which is appropriate since UpSet breaks on empty data.

### DIFF
--- a/src/candela/components/UpSet/index.js
+++ b/src/candela/components/UpSet/index.js
@@ -55,6 +55,9 @@ export default class UpSet {
     if (!this.options.id || (!this.options.sets && !this.options.fields)) {
       return;
     }
+    if (!this.options.data || this.options.data.length === 0) {
+      return;
+    }
 
     d3.select(this.el).html(template);
 


ronichoudhury:
LGTM.

alex-r-bigelow:
@ronichoudhury: should I start a stable-filter-interface branch for the vagrant script (where we could merge this change)?

If not, you'll probably want to stick with the 9d7b9f58a6a157a9ac41488f4a85aa5bd1f25024 commit that we're already using (the latest commit is a messy WIP)

ronichoudhury:
I spoke to Alex off-line about a "stable-filter-interface" branch and I
think it's a good idea. I'll update the Ansible provisioning to point to
this branch when it's ready.

jeffbaumes:
Since this is not a messy change I assume I can merge to master, and you would also merge it to stable-filter-interface? I'll wait for confirmation before merge to master.

ronichoudhury:
Yes, sorry, I should have been more clear. Go ahead and merge to master
now, and in the meantime Alex will create a semi-stable branch containing
his changes, and we can continuously merge master to it for the purposes of
keeping the NCI demo up to date.

PR #250, opened 2016-05-23
Add a default.nix file to enable building in NixOS environments

ronichoudhury:


ronichoudhury:
Since this is useful to me and probably me only, and it doesn't actually touch any project code, I'm going to self-LGTM this and merge.

PR #256, opened 2016-06-03
Matching makeover

alex-r-bigelow:
In addition to various color / style changes, this PR adds bubbles to the matching interface that indicate:
1. how many links a vis encoding can accept
2. how many links have been established
3. whether or not the encoding is required
4. whether or not the requirement has been satisfied
The indicator at the top of the matching interface consequently makes more sense

Also, an additional "click to delete" feature was added to links


alex-r-bigelow:
Just realized there are a few bugs... will remove the WIP shortly.

jeffbaumes:
This looks really slick!

It would be good if the links had a bit larger target for hovering. This can be done by including an additional transparent link with larger stroke width. Also would be good to have a helpful tooltip when hovering over the link that states that the link will be deleted, maybe "Click to remove this matching"?.

jeffbaumes:
I also notice that bubble sizes aren't consistent. Anything we can do to make them always the same size? If we want them to grow when there are 10 or 100 connections to a repeatable option, that would be ok, but it seems we could have a minimum size for when they there is only one digit so that "1/1" is not smaller than "0/1".

jeffbaumes:
I think there should be a max-central-gap so that this doesn't happen on large screens:

screen shot 2016-06-03 at 12 57 24 pm


Maybe push things toward the center in cases like this?

jeffbaumes:
Tiny thing (🐜 ): "0 / ∞" still seems to push the bubble a tad bigger for me. Can you make the min size just a bit larger?

alex-r-bigelow:
Okay, I think I've addressed all @jeffbaumes's comments... let me know if there's anything else. Thanks for taking the time to review!

ronichoudhury:
This looks awesome! I have a few interface issues to discuss but it's nothing that should stop this from being merged. LGTM.

PR #257, opened 2016-06-03
GeoJS and GeoDots components

ronichoudhury:
This introduces `Geo`, a generic component meant to produce GeoJS plots, and `GeoDots`, a thin wrapper around Geo that accepts ready-made location data in order to plot points on a map.

The `spec` getter has been implemented, so you can load up e.g. the airports data set in Resonant Laboratory and attach lat and long to y and x and see dots on a map.


alex-r-bigelow:
This is **so** cool! 

I'm getting a bunch of console errors along the lines of:
```
d3.js:670 Error: <circle> attribute cy: Expected length, "NaN"
```
when the project doesn't have any matchings set up yet.

This may be a more general Resonant Lab problem... should we even try to render a visualization when we already know that the required matchings aren't satisfied—in this case, would some generic message be better than showing an incomplete visualization?

alex-r-bigelow:
I also seem to be getting
```
geo.js:10352 Uncaught Error: Invalid viewport dimensions
```
when I resize the window.

jbeezley:
That typically means the height of the element geojs is placed in is 0 (when the resize event is triggered).  You either need to give that element an explicit height in CSS or turn off [auto resizing](https://github.com/OpenGeoscience/geojs/blob/master/src/map.js#L60) and manage the resize events manually.

alex-r-bigelow:
@manthey and I just probed this a little, and it looks like this is a result of Resonant Lab trashing the GeoDots component and starting fresh when it can't figure out how to pass it (potentially) new data... essentially, when the old component is removed, the container shrinks to size 0, and the old component naturally complains about it.

So there are two solutions: like @jbeezley mentioned, we can turn off auto resizing **or** the problem will magically disappear on its own if we add an update function to the GeoDots component.

src/candela/components/Geo/index.js
```diff
@@ -0,0 +1,39 @@
...
+    super(el);
+
+    const geojs_opt = options.geojs || {};
+
+    this.plot = geojs.map(Object.assign({
+      node: el
+    }, geojs_opt));
+
+    this.plot.createLayer('osm', {
+      renderer: null
```
alex-r-bigelow: @manthey mentioned that this may not be the best choice: I think we can tell GeoJS that it's a point layer, and it will figure out the best renderer for us.

ronichoudhury: Not sure I understand - before creating a point feature we need to create a layer, right? Perhaps @manthey can set me straight.

In any case, the layer should not be created this way in the constructor, which is an oversight on my part. I moved other construction-time options into that `geojs_opt` mechanism and I think we need something similar for creating the base layer. I think for now just setting another top-level option called `baseLayer` or something like that will suffice. I'll make the change and then collect feedback from the GeoJSes :smile_cat:

ronichoudhury:
I think the best way forward is to come up with a general way for a component to have an `update()` method. It would be optional and Resonant Lab could use it if present to update a vis in place (as it looks like we need to for the GeoDots component). This component could use such a method, for a better reason than errors appearing in the console: every time the window resizes, the map is torn down and rebuilt using the original `center` property, whether explicitly stated or not. This means the map will, in general, jump somewhere other than where the user was, every time the map is rebuilt.

See issue #258

alex-r-bigelow:
With a little more experimentation, I'm pretty sure that both errors that I'm seeing are a symptom of the same problem. As that will need its own PR, this one LGTM.

ronichoudhury:
@aashish24, @jeffbaumes, @alex-r-bigelow: this is ready for final review.

jeffbaumes:
Looks like you need to merge in master first due to conflicts. Can you do that and then we can do final review?

ronichoudhury:
Sure. I'll rebase against master and force push, if that's ok with you.

jeffbaumes:
Works for me thanks.

ronichoudhury:
I rebased and force-pushed, FYI to everyone.

I'm also adding a couple of bugfixes that Jeff requested, so hold off on reviewing until I'm done with those.

ronichoudhury:
Whew ok, this should now be ready for final review!

alex-r-bigelow:
Tiny :ant: comment / question: it looks like the tile data attribution has a `z-index` of 1001, and this makes it float on top of Resonant Lab's overlays and menus (e.g. try the dataset menu). Resonant Lab's z-indices are more on the order of 100-110. I'm more than happy if we just file an issue to boost Resonant Lab's z-indices a digit instead—this doesn't have to affect this PR.

Other than that and the image tests, this LGTM!

jbeezley:
That's the kind of thing that I would like to make configurable in https://github.com/OpenGeoscience/geojs/issues/431 when we get around to doing it.

ronichoudhury:
@alex-r-bigelow what would be a good z-index to use for the attribution layer? We can set an `!important` rule in Resonant Lab's CSS.

ronichoudhury:
A value of 1 seems to do it (keeps it on top of the map but under ResLab UI). I'll push and you can test it out Alex.

alex-r-bigelow:
I tend to shy away from `!important` when I can help it—I think that's a sign that the correct way to do this is indeed to boost Resonant Lab's `z-index` instead.

ronichoudhury:
Should I just add 900 to each of ResLab's layer z-index values?

jbeezley:
GeoJS actually uses `z-index` properties to reorder layers.  So if you ever call `layer.moveUp` or add a UI layer, then you will get a larger `z-index`.  I think I set it up so the layer `z-index` will always be less the number of layers.

ronichoudhury:
@jbeezley good to know. We instead boosted the z-index of the relevant parts of Resonant Lab and it seems to be working now without poking at GeoJS policies.

ronichoudhury:
@jeffbaumes @aashish24 PTAL.

jeffbaumes:
LGTM!!

PR #259, opened 2016-06-06
Add branch-based versioning of projects to help us sandbox better as we develop

### LINKED ISSUE #255: Store / filter items based on Resonant lab version numbers
We run into this problem a lot: a given dataset or project item won't work with whatever version of resonant lab is being used.

Two ideas:
- For broader deployment, we should store Resonant Lab version numbers as metadata in these items, so we can update old items' metadata if/when necessary. Users should never have to deal with file version issues (MS Office, anyone?)
- For internal development, we could instead store the current branch name as the version (or commit hash?) via some kind of `subprocess.check_output(["git", "status"])` voodoo, so that 1) we don't run into bugs that have nothing to do with the current version that we're working on, and 2) we don't accidentally corrupt dataset/project metadata that was supposed to be stable. In development mode, we'd hide any items in the interface that don't match the current branch

alex-r-bigelow:
Fix #255

Projects are hidden in menus if they were created on a different branch than the current one (all projects are still accessible by URL)

### DIFF
--- a/app/resonant-laboratory/mainPage.js
+++ b/app/resonant-laboratory/mainPage.js
@@ -23,6 +23,17 @@ window.girder.apiRoot = 'api/v1';
 // Our main view that coordinates each big chunk
 let MainPage = Backbone.View.extend({
   initialize: function () {
+    // Get the current app version
+    new Promise((resolve, reject) => {
+      window.girder.restRequest({
+        path: 'system/resonantLaboratoryVersion',
+        type: 'GET',
+        error: reject
+      }).done(resolve).error(reject);
+    }).then(versionNumber => {
+      this.versionNumber = versionNumber;
+    });
+
     // Set up navigation
     this.router = new Router();
 
--- a/app/resonant-laboratory/models/Project.js
+++ b/app/resonant-laboratory/models/Project.js
@@ -24,6 +24,7 @@ let Project = MetadataItem.extend({
     return {
       name: 'Untitled Project',
       meta: {
+        versionNumber: window.mainPage.versionNumber,
         datasets: [],
         matchings: [],
         visualizations: [],
--- a/app/resonant-laboratory/server/__init__.py
+++ b/app/resonant-laboratory/server/__init__.py
@@ -1,6 +1,7 @@
 import os
 from girder.api.rest import Resource
 from anonymousAccess import AnonymousAccess
+from versioning import Versioning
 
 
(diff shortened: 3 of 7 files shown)


ronichoudhury:
How much extra work would it be to *mark* matching projects with a green checkmark, etc., and non-matching ones with a red X, rather than simply hide all non-matching ones? I ask because I can imagine this causing the opposite problem to the one it solves - someone switches to a new topic branch and suddenly they can't see any projects, even ones that will work because the metadata schema isn't going to change on that topic.

ronichoudhury:
Also, what happens if a user doesn't have `git` on their path (because, e.g., it isn't installed)?

alex-r-bigelow:
It would be a little bit of work... actually, it would make sense to include a (tiny) project version compatibility indicator as part of the glyph redesign in #164. For now, to use an old project on a new branch, the URLs will still work for any project—they just won't show up in the menu.

For now, the "version" comes back as `master` if the branch is actually master, release, or if `git status` fails for whatever reason (including the case where git isn't installed). In the future, I'm imagining an actual number to come back in these three scenarios.

ronichoudhury:
Handling of nonexistent `git` sounds good.

Alex and I just discussed in meatspace, and he'll put a crude stopgap version compatibility indicator into the project icons as part of this PR, then rework that as necessary as part of #164. When that's done I can review one more time (since the code as is looks good and I'm just concerned about unconditionally hiding non-matching projects).

alex-r-bigelow:
Turns out this wasn't as tricky as I thought. Great idea!

ronichoudhury:
Very nice!

Do you think it makes more sense to mark the non-conforming projects with a warning, or the conforming ones with a green OK sign? I would think green OK sign because that will be the thing that draws your eye, but since you're getting the most benefit out of this feature I'll defer to you. Just let me know so I can ok this branch.

alex-r-bigelow:
Ideally, a deployed app shouldn't have to worry about displaying any version information (we'd want to do something more formal like @jeffbaumes proposed to gracefully handle forward/back conversion)—a deployment of `master` or `release` should never contain our weird branch projects in the first place. My thoughts are that this indicator adds clutter that we really only care about as we're developing it. So for now I think it's best to just show a warning when there's a potential version conflict.

ronichoudhury:
Oh ok I see. You wouldn't want all projects showing up with the green indicator on master.

Sounds good! LGTM.

PR #263, opened 2016-06-09
Branch sandboxing

alex-r-bigelow:
This fixes a couple a couple small versioning bugs that I encountered when re-deploying to my Utah server


ronichoudhury:
LGTM

PR #264, opened 2016-06-13
Visualization serialization

jeffbaumes:
This is an initial serialization prototype. `getSerializationFormats()` provides a list of serialization types, which may be sent to `serialize()`.

The examples page will produce buttons for downloading each serialization type supported by the visualization. Currently only works with the scatterplot example.

### DIFF
--- a/src/candela/components/ScatterPlot/index.js
+++ b/src/candela/components/ScatterPlot/index.js
@@ -85,6 +85,19 @@ export default class ScatterPlot extends VisComponent {
   }
 
   render () {
-    vega.parseChart(spec, this.el, this.options);
+    this.chart = vega.parseChart(spec, this.el, this.options);
+  }
+
+  getSerializationFormats () {
+    return ['png', 'svg'];
+  }
+
+  serialize (format) {
+    if (!this.chart) {
+      this.render();
+    }
+    return this.chart.then((vobj) => {
+      return vobj.toImageURL(format);
+    });
   }
 }
--- a/src/candela/components/ScatterPlot/index.js
+++ b/src/candela/components/ScatterPlot/index.js
@@ -1,8 +1,8 @@
 import VisComponent from '../../VisComponent';
-import vega from '../../util/vega';
+import VegaChart from '../../VisComponent/mixin/VegaChart';
 import spec from './spec.json';
 
-export default class ScatterPlot extends VisComponent {
+export default class ScatterPlot extends VegaChart(VisComponent, spec) {
   static get spec () {
     return {
       options: [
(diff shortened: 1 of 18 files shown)


src/candela/components/ScatterPlot/index.js
```diff
@@ -85,6 +85,19 @@ export default class ScatterPlot extends VisComponent {
...
+
+  getSerializationFormats () {
+    return ['png', 'svg'];
+  }
+
+  serialize (format) {
+    if (!this.chart) {
+      this.render();
+    }
+    return this.chart.then((vobj) => {
```
ronichoudhury: I'd say get rid of the parentheses around `vobj` (unless we decided on a style that keeps them for single-argument arrow functions).

src/candela/components/ScatterPlot/index.js
```diff
@@ -85,6 +85,19 @@ export default class ScatterPlot extends VisComponent {
...
+    this.chart = vega.parseChart(spec, this.el, this.options);
+  }
+
+  getSerializationFormats () {
+    return ['png', 'svg'];
+  }
+
+  serialize (format) {
+    if (!this.chart) {
+      this.render();
```
ronichoudhury: I don't think serializing should cause a render. In this case I'd prefer to return a fulfilled promise that delivers `undefined` or `null` or a blank image.

ronichoudhury:
Very cool!

jeffbaumes:
@ronichoudhury PTAL. I've removed parens and changed serialize to return a rejected promise, which is tested in the example index.js - I ensure that the promise is rejected if render has not been called.

ronichoudhury:
Rejected promise! Great idea :tada:

ronichoudhury:
One thing I'm wondering about: since we're using a getter to store the options spec for a `VisComponent`, should we do something similar for serialization formats? It seems like they should either both be getters, or both be methods. Thoughts?

(Aside from this discussion, the code looks pretty good, FYI.)

jeffbaumes:
Good point about getters. Instead adding another field to `spec`, I propose we get rid of `spec` entirely and have:

```
static get options()

static get container()

get serializationFormats()
```

`options` and `container` need to be static since you need them to initialize the visualization, while `serializationFormats` does not need to be static, so probably shouldn't be. Also the formats could in theory vary based on how the vis is initialized.

ronichoudhury:
I like the options/container/serializationFormats getters proposal.

jeffbaumes:
This is ready for final review. I think splitting spec into options/container should go in another PR.

ronichoudhury:
The `VegaChart` mixin looks awesome!

If we can just fix the problem with the UI buttons spoiling the image tests and Travis is otherwise happy, then LGTM :smiley_cat:

jeffbaumes:
I made Travis happy so I'm merging due to @ronichoudhury's provisional LGTM.

ronichoudhury:
:fireworks:

PR #268, opened 2016-06-16
Eliminate artificial 1-second delay in image tests

ronichoudhury:


jeffbaumes:
Travis approves, so me too. LGTM! 🏃

Issue #262, opened 2016-06-07
Reduce delay in matching panel animation


ronichoudhury:
Opening the matching panel, or causing it to resize and/or move (by manipulating the neighboring panels) reveals a delay before the matching panel contents re-render.

There is also a bit of a delay after clicking on an element of the matching interface, but I don't know if it arises from the same source as the re-render delay.

There is also a "two-phase" animation that happens when the panel size changes; the boxes show the animation delay while the lines pop to where they are supposed to end up. The "second phase" seems to be a bug in animation, perhaps related to the same re-render delay problem described above.

PR #270, opened 2016-06-21
Adding export button to vis panel

jeffbaumes:
Left to do (that I'm hoping @alex-r-bigelow will help with)

 * [x] Create export icon in vis title bar (currently a green checkmark)
 * [x] Create download icons in export overlay
 * [x] When the vis panel is collapsed, the button doesn't work (I think the vis is undefined). Simplest seems to be to remove the export option when the view is collapsed.


alex-r-bigelow:
Yes, to avoid unnecessary (and probably buggy) render calls to vis components when their container will have a size of zero, the component is destroyed altogether when the widget is collapsed.

Idea 1: I wonder if it would be better to just move the containers offscreen when the widgets are collapsed—this would do a few things. Any state inside the vis component would be preserved (e.g. geo viewport, UpSet choices of which sets to show, etc). It would also make it easy for us to specify an image size in the download, instead of just using the size that the vis happens to be in the browser (probably an icebox feature).

Idea 2: I've played a little with just forcing the vis panel open when the export button is clicked, but it looks like there are several debouncing / timing problems. Potentially, the vis component potentially might not even get instantiated until after a client-side dataset parse (or in the future, pulling a page of items from the server). This isn't a big deal for the actual serialization (vis shouldn't be undefined anymore in the time it takes the user to click the button), but to get at `serializationFormats` in the first place, this is a problem. Maybe we could just access `serializationFormats` statically from candela?

Idea 3: We could add stuff to disable the export button unless the view is open. The problem here is we don't have any code / design for disabling widget buttons yet. Weirdly, idea 1 may be less work.

jeffbaumes:
Idea 3 seems much simpler to me. Couldn't the lines [here](https://github.com/Kitware/candela/blob/reslab-export/app/resonant-laboratory/views/widgets/VisualizationView/index.js#L96-L98) just be changed to

```
if (!widgetIsShowing) {
    this.vis = undefined;
    // Remove Export button
} else {
    // Add Export button
}
```

On a side note, there has been a time or two where I wished `serializationFormats` was static also, so that is worth considering.

alex-r-bigelow:
It gets a little messier than that... the widget panel stuff (including the icons) are rendered before the actual widget itself. So you'd have to do a double render to actually get the icon to show up + something to prevent an endless render loop. Or we could insert the icon directly in VisualizationView's render call, but that would duplicate code—not to mention messing with the same piece of the DOM from two totally different places.

In the long run, I imagine we'd want Idea 1 anyway (if we do it right, it could help fix #260 and #262)... it may be as simple as a couple CSS tweaks. I'll take a crack at it this morning.

alex-r-bigelow:
@jeffbaumes I think this is ready.

jeffbaumes:
@alex-r-bigelow's changes/icons LGTM. I talked to @ronichoudhury and he's ok with merging even though there is a spurious failure on heatmap that is unrelated to this branch.

PR #272, opened 2016-06-22
Allow for multiple baseline images in image testing

ronichoudhury:
To use this:
1. Check a new baseline image into the testing directory. By convention, it should be named after the test itself, followed by special description. e.g., `heatmap-travis.png` describes a baseline for the Heatmap component that is meant for use in Travis testing.
2. Add an `extraBaselines` parameter to the appropriate invocation of `imageTest()`. This property should be a list of basenames. For our example, it would be `extraBaselines: ['heatmap-travis']`.

During the image test, the test image will be compared to each of the baselines (including the implicit baseline formed by appending `.png` to the test name); the smallest error value will be used as the basis for the test's pass/fail status.


jeffbaumes:
LGTM

PR #275, opened 2016-06-24
[WIP] Custom filesystem assetstore adapter

ronichoudhury:
**Note:** *I am not currently planning to actually merge this to master; it's here mainly to keep everyone in the loop and serve as a point of discussion.*

This PR engages Girder's new capability to install custom classes into its factory for assetstore adapters. The one implemented here falls back on the usual adapter in most cases, and exhibits new behavior if `extraParameters` is set.

Right now, as a matter of proof of concept, it accepts two parameters: `offset` and `limit`, and the target file/item must be a CSV file. The adapter extracts the header line and then `limit` data lines, beginning with the line numbered `offset`.

Generally speaking, that behavior will be replaced with more careful behavior that can understand Mongo-style database filter queries in addition to limit/offset, and also returns the data in a JSON-objects style (and of course this behavior can change to suit whatever is needed by Resonant Laboratory).

I have already discussed some code strategy with @alex-r-bigelow. He has two branches in-flight with a relatively large set of changes; once those are stable and merged together, my idea is that he can merge this branch into his, and from there we can work on the specific semantics we want in earnest.

I have also discussed with @jeffbaumes the idea that this adapter extension maybe ought to exist more as a decorator for existing classes, rather than a subclassing of a specific adapter class, so that its behavior can be used with both `FilesystemAssetstoreAdapter` and `GridFSAssetstoreAdapter`.

### DIFF
--- a/app/resonant-laboratory/models/Dataset.js
+++ b/app/resonant-laboratory/models/Dataset.js
@@ -115,6 +115,7 @@ let Dataset = MetadataItem.extend({
           data: {
             extraParameters: JSON.stringify({
               fileType: 'csv',
+              outputType: 'csv',
               offset: 47,
               limit: 5
             })
--- a/app/resonant-laboratory/models/Dataset.js
+++ b/app/resonant-laboratory/models/Dataset.js
@@ -114,6 +114,7 @@ let Dataset = MetadataItem.extend({
           path: 'item/' + this.getId() + '/download',
           data: {
             extraParameters: JSON.stringify({
+              fileType: 'csv',
               offset: 47,
               limit: 5
             })
--- a/app/resonant-laboratory/models/Dataset.js
+++ b/app/resonant-laboratory/models/Dataset.js
@@ -112,6 +112,12 @@ let Dataset = MetadataItem.extend({
       }, () => {
         return girder.restRequest({
           path: 'item/' + this.getId() + '/download',
+          data: {
+            extraParameters: JSON.stringify({
+              offset: 47,
+              limit: 5
+            })
+          },
           type: 'GET',
           error: null,
           dataType: 'text'
--- a/app/resonant-laboratory/server/SemanticFilesystemAssetstoreAdapter.py
+++ b/app/resonant-laboratory/server/SemanticFilesystemAssetstoreAdapter.py
@@ -1,6 +1,6 @@
+import cherrypy
 import csv
 import json
-import os
 import sys
 from girder.models.model_base import GirderException
 from girder.utility.filesystem_assetstore_adapter import FilesystemAssetstoreAdapter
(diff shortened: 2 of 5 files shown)


alex-r-bigelow:
This looks pretty awesome!

I'm curious about "mongo-style" filter queries... I assume this refers to the slimmed-down syntax we discussed along the lines of:
```
{
  "left": { ... nested expression ... },
  "operator": "notin",
  "right": { ... nested expression ... }
}
```
Or were you thinking of actually using mongo syntax directly (something like [sift.js](https://github.com/crcn/sift.js), but in python)? If so, it might make more sense for me to just use mongo filter syntax directly in the client.

ronichoudhury:
Something like sift would be fine, though I don't think we need to immediately support everything it supports. We really just need `$and` and `$or` I think, at least for now. We don't have to use the `left`/`operator`/`right` approach either; it could just be two expression types: `{'$and': [...]}` and `{"$or": [...]}` but my point is, we are free to use whatever we think is easiest/best.

alex-r-bigelow:
By following mongo's syntax, we could potentially articulate more advanced queries that our slimmed-down syntax wouldn't support. This has pros and cons:

Pros:
- Uses an existing standard, rather than us having to make up (and convert!) our own filter syntax
- Way down the road, we've talked a little about visualizations themselves specifying their own advanced filters (e.g. a candela component being part of the query interface instead of the end visualization). If we're already passing queries around in mongo format, we wouldn't have to do any refactoring or conversion to add more advanced filtering capabilities—they'd be supported as soon as they're implemented for each assetstore.

Cons:
- Using mongo syntax may imply that we support more than we actually do—like your comment, we'll have to be clear in the documentation that we only support `{$and: [...]}` and `{$or: [...]}`
- Would `{$and: [...]}` and `{$or: [...]}` be any trickier to implement (probably not, but this is up to you)?
- I don't know the plans with regard to the filter syntax in girder_db_items: giving mongo syntax first-class status in Resonant Lab/Candela could mean trickier conversions to something like SQL somewhere down the road. If we only support `$and` and `$or`, this may not be a big deal (with the current girder_db_items syntax, I think only `$and` is supported anyway?), but if we start getting more advanced, things could get messy really quickly

I think my vote would be to use mongo-style `{$and: [...]}` and `{$or: [...]}`, but I just want to make sure that's not going to be harder to implement / cause problems with girder_db_items.

jeffbaumes:
+1 to mongo subset, but I'd like @manthey to chime in also when he's back around.

ronichoudhury:
I think the cons are less conny than might appear. "Support more than we do" isn't a big problem in my view. Documentation will help, and we may opt to use `and` and `or` rather than the `$`-prepended variants, just to draw the distinction (conversion to a PyMongo query would then remain trivial).

Using the list syntax for `$and` and `$or` will be marginally more difficult to implement, depending on what we're talking about exactly. If we're using PEGjs to parse user-generated filter expression strings, then there's one extra step to turn a binary tree into a rose tree and I don't think it's a big deal. But that's the hardest problem with the list syntax I can think of.

I'm not too worried about supporting SQL queries in the future. From a compound expression consisting only of `and` and `or` queries, there are standard ways to express the equivalent in SQL. We would just need to figure out the right way to do so, then generate the SQL query string that would be sent to the server. @manthey is the resident expert on that so I'm hoping he can confirm my sunny outlook.

manthey:
A couple of thoughts on this:

In the current girder_db_items implementation, I'm not using Mongo's
format.  We were asked to support GIS functions in Postgres, which I did by
allowing functions to be used in any place where a field (column) name
could be specified.  This can be done in the query, in the fields list (the
projection in Mongo), or the sort.  Mongo doesn't specifically have this
functionality -- you can use some functions in a $where clause, but this is
fairly different, and I don't think you can use functions in the projection
or sort.

At present, we haven't had any requests to support Mongo GIS queries (which
are done using special $ operators rather than functions).  We could make a
format that is Mongo-like where Postgres functions are also prefixed with
$, or add a special $func operator, or ...

This means that either (a) we support both a Mongo-subset AND our own
format, (b) we support a Mongo-subset with extensions, or (c) we just use
our own query format.

So, to sort-of repeat what Alex said, using a Mongo-subset with extensions:

Pros:
- For simple cases, uses an existing standard
- Fairly straightforward to add more features and support.

Cons:
- We will only support a subset of the standard
- We have to have our own extensions to handle non-Mongo database features,
which makes those queries possibly more complicated.
- Sorts and projections need to have special extensions to handle functions

David


On Fri, Jun 24, 2016 at 2:11 PM, Roni Choudhury <notifications@github.com>
wrote:

> I think the cons are less conny than might appear. "Support more than we
> do" isn't a big problem in my view. Documentation will help, and we may opt
> to use and and or rather than the $-prepended variants, just to draw the
> distinction (conversion to a PyMongo query would then remain trivial).
>
> Using the list syntax for $and and $or will be marginally more difficult
> to implement, depending on what we're talking about exactly. If we're using
> PEGjs to parse user-generated filter expression strings, then there's one
> extra step to turn a binary tree into a rose tree and I don't think it's a
> big deal. But that's the hardest problem with the list syntax I can think
> of.
>
> I'm not too worried about supporting SQL queries in the future. From a
> compound expression consisting only of and and or queries, there are
> standard ways to express the equivalent in SQL. We would just need to
> figure out the right way to do so, then generate the SQL query string that
> would be sent to the server. @manthey <https://github.com/manthey> is the
> resident expert on that so I'm hoping he can confirm my sunny outlook.
>
> —
> You are receiving this because you were mentioned.
> Reply to this email directly, view it on GitHub
> <https://github.com/Kitware/candela/pull/275#issuecomment-228419350>, or mute
> the thread
> <https://github.com/notifications/unsubscribe/AIX_RyBfizNXV_53g77TdveMoPbTxiEqks5qPB3sgaJpZM4I9zIl>
> .
>



-- 
David Manthey
R&D Engineer
Kitware Inc.
(518) 881-4439

ronichoudhury:
As of 4cfae3cae14f4bd4e2fe1acaf8e3e66a226a0473, this branch has been merged into filter-makeover, so any further work that needs to be done should be in the form of PRs to that branch.

PR #276, opened 2016-06-27
Update build docs for Resonant Lab

danlamanna:
@alex-r-bigelow

### DIFF
--- a/app/resonant-laboratory/README.md
+++ b/app/resonant-laboratory/README.md
@@ -13,6 +13,7 @@ Choose your own visualization adventure.
   ```bash
   cd candela_repo_dir
   npm install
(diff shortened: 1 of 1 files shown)


alex-r-bigelow:
LGTM, thanks!

ronichoudhury:
Thanks for the PR @danlamanna.

What happens if you don't do the `npm run build` step? I didn't think the explicit build of the Candela library was needed to build Resonant Laboratory.

ronichoudhury:
Ah, ok yeah there's a special one-off phase where LineUp is built in order to circumvent some problems with a release build. I didn't realize it would cause problems for ResLab. Thanks again!

PR #301, opened 2016-07-15
Implement handling of not-clauses in Mongo backend

ronichoudhury:
The `not` keyword was a problem for Mongo due to the way it works in Mongo queries. The solution in this PR is to make a pass over an AST to remove all instances of `not`-clauses. This is done by replacing the `not`-node with an appropriate node that implements the inverse semantics. For example, `not(and(age > 5, age < 7))` becomes `or(age <= 5, age >= 7)` (invoking DeMorgan's laws to make the transformation).

This also adds a bit of testing for `not`-expressions.

### DIFF
--- a/app/resonant-laboratory/querylang/test/query.js
+++ b/app/resonant-laboratory/querylang/test/query.js
@@ -20,6 +20,18 @@ const disjunction_expression = [
   'age > 22 or age < 20'
 ];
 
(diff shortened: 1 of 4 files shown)


alex-r-bigelow:
Sorry for the delay... somehow I must have missed the notification that this PR was up. LGTM!

Issue #284, opened 2016-07-12
Add filtering functionality to histograms


alex-r-bigelow:

alex-r-bigelow:
Fixed in #316

Issue #307, opened 2016-07-19
Query language does not handle type coercion


alex-r-bigelow:
Example of the problem:
- The filters are created in terms of the user's type coercion settings. Meaning that `cylinders not in [4,5,6]` is expecting the data to already be a number... but the filter is applied in a setting that is unaware of type coercion parameters (e.g. `astToFunction` produces something that operates in python land, iterating over what are likely rows of CSV strings).

- We need to do one of the following:
1) Somehow make type coercion a native part of the query language
2) Somehow reach into whatever process is actually handling the data, and convert values on the fly before they reach the query function / mongo query
3) Actually go in and change the native values in the database / flat file per the user's coercion settings

1 and 3 may or may not be possible / practical (@ronichoudhury, thoughts on this?). But the problem with 2 is it requires adding a lot of app-specific logic in places that it probably shouldn't exist, such as providing a custom preprocessing function to the download endpoint as an `extraParameter` before the filter is applied... and that function would somehow have be executed in python for flat files, as well as inside mongodb, etc.

Thoughts on how we should proceed? I'm leaning toward 1 if it's possible—that seems the cleanest to me.

Minor, but related points:
- We have semi-specialized coercion logic in `general_purpose/binUtils.js`, e.g. we use Javascript nuances like truthy/falsey values for boolean conversion. Ideally, we'd like to avoid duplicating this code with all its nuances, but that may be unavoidable.
- Our decision here will probably determine a lot about how we handle corner case values like `"cylinders" not in [NaN]`

jeffbaumes:
I think that 3 is out - the data should just be the data and the coercions should happen on-the-fly. The other two options are basically variations on how to get the coercion settings over to the server. 2 does not sound great since the general data extraction should not need to know about coercion.

1 is better, but another way to solve it is to send the coercion data separately - it does not necessarily need to be part of the query language. Could something simple like an extra argument with the information like `{ "cylinder": "integer", ... }` work?

alex-r-bigelow:
The problem is, where would that argument be implemented? If I understand it correctly, the plan is for data extraction to accept a filter parameter. Because coercion has to happen *before* the filter is applied, data extraction would also have to accept (and implement the logic for) that extra argument. If we don't want that, the only other way I can see us sneaking it in would be as part of the query language.

alex-r-bigelow:
If none of these are palatable, I guess there's one more option:
[...]

PR #309, opened 2016-07-20
[WIP] Filter interface

### LINKED ISSUE #168: Filtering interface / endpoints

alex-r-bigelow:
Adds filtering capabilities to Resonant Laboratory

Fixes #168, #284


alex-r-bigelow:
At this point in time, this PR includes all the client-side basics of the filter interface, with histogram calculations function correctly for mongo-based datasets.

The remaining functionality (mapreduce for non-mongo datasets, pulling the correct data in the table) depends on a download endpoint that accepts a filter parameter, i.e. we are waiting on [this](https://github.com/OpenGeoscience/girder_db_items/pull/8) PR to be merged.

I will start a separate `filter-interface-download` branch that will refactor resonant lab to make use of the new girder_db_items plugin, and merge it into this PR once girder_db_items is ready.

app/resonant-laboratory/grunt-watch-plugin.js
```diff
@@ -12,7 +12,7 @@ module.exports = {
   apply: function (compiler) {
     compiler.plugin('emit', (compilation, callback) => {
       if (compilation.options.watch === true) {
-        exec('cd $GIRDER_PATH && grunt', function (error, stdout, stderr) {
+        exec('cd $GIRDER_PATH && grunt --force', function (error, stdout, stderr) {
```
ronichoudhury: What's the reason for adding `--force` here?

alex-r-bigelow:
At this point, this branch requires two (possibly three) major changes before it's eligible to be merged into master:
- [ ] We need `semantic_access()` to support DB assetstores—mongo-based datasets will show a broken table view / connected visualizations will not work until this is fixed
   - [ ] Probably would use the same logic as streaming JSON files?
- [ ] Need support for type casting in queries (#307)—csv-based datasets will have odd/incorrect filter behavior until this is fixed

@ronichoudhury, I'm happy to work on either of these; I just want to make sure I won't be stepping on your toes if I proceed.

app/resonant-laboratory/grunt-watch-plugin.js
alex-r-bigelow: For some reason grunt is tripping up (and crashing) on the `database_assetstore` plugin. It's possible there's something weird in my girder installation as a result of switching branches... I'll try a fresh girder installation and see if this problem goes away.

package.json
```diff
@@ -9,7 +9,7 @@
     "build:laboratory": "cd app/resonant-laboratory && webpack",
     "build:laboratory:tests": "webpack --config ./app/resonant-laboratory/webpack-testing.config.js",
     "test:laboratory:unit": "npm run test:laboratory:unit:raw | tap-spec",
-    "test:laboratory:unit:raw": "node build/resonant-laboratory/tests.js",
+    "test:laboratory:unit:raw": "for t in `ls build/resonant-laboratory/*.tests.js`; do node $t; done",
```
ronichoudhury: What was the reason for this change?

alex-r-bigelow: I added a few of my own tests; if there's a better way to do this, I'm happy to change it.

ronichoudhury: Oh I see. I think you can pass multiple files directly to `node` but maybe that's not right.

This isn't that big of a deal, it's just that it depends on a unix-type shell to work properly.

app/resonant-laboratory/grunt-watch-plugin.js
ronichoudhury: In that case I'd prefer we not just put `--force` in here. Let's try to figure out what's going wrong on your machine first.

alex-r-bigelow:
To replicate both behaviors:

1) Run the new `scripts/populateGirder.py` script—this will upload a bunch of flat CSV files and mongo db files

2) Create a new project with the `cars` dataset (this one is a CSV file)
  - The filters should misbehave a bit—you can inspect the filters that the interface is generating by inspecting the `getHistograms` requests. Or you can see the human-friendly version of the filters by typing (Chrome will probably offer to autocomplete the right girder ID):

    `mainPage.loadedDatasets[' --- the dataset's girder ID --- '].listAllFilterExpressions()`

    or to see the resulting AST tree:

    `mainPage.loadedDatasets[' --- the dataset's girder ID --- '].getFilterAstTree()`

3) Switch to the `airports` dataset (this one is a mongo collection)
  - Here the filters should work as expected
  - The table view should be broken; if you inspect the `download?extraParameters=...` request, you'll see the kind of thing that comes back from the current download endpoint. This *should* look a little different:
    - The download endpoint needs to use the filter / limit / offset parameters that we're sending it
    - We should be getting the contents of the `data` property without the meta information (not a big deal; we can always do a check and pull it out if that's what's supposed to come back from downloading a database table)

alex-r-bigelow:
I accidentally closed this PR... because of difficulties in reopening it, we have moved it to #316

PR #310, opened 2016-07-20
SurvivalPlot component

ronichoudhury:
This introduces a `SurvivalPlot` component. The new example app demonstrates how it works.


alex-r-bigelow:
This is neat; it LGTM.

If we start getting a lot of specialized candela components like this (e.g. this is *almost* the same thing as the line chart, but with some secret sauce), is there a plan for categorizing them? I'm mainly asking for the sake of a future Resonant Lab menu redesign with unique glyphs for each candela component (#164). Is there a way that we'd want candela components to describe themselves in the spec (e.g. point to an image as its icon)?

ronichoudhury:
I haven't thought about categorizing yet but it's definitely something we should consider. I want to be careful about two things: (1) designing too closely to what Resonant Lab wants (by, e.g., requiring a lot of stuff that's meant solely for UI reasons), and (2) lining up a graphical design workload (for icons, etc.) in the future that grows linearly with the number of components.

On the other hand, since Resonant Lab may be the main way new users come to Candela, some amount of UI design "sparkle" may be helpful.

jeffbaumes:
I like the self-contained idea of having the icon alongside the component in Candela (I think I brought it up at one point) but I also see Roni's point. There certainly shouldn't be a requirement that a Candela component needs an icon to be included in Candela - there should be a generic icon. I could go either way on whether the icons we do have should be in Candela or ResonantLab.

As for categorization, I think it's a good idea but not for the sake of icons. I don't think I'd want something like survival plot to be part of a "Line Charts" category. Instead, specialized vis should be categorized more based on data type or domain than on what it ends up looking like. A "Clinical" category might be a good place to put this plot.

alex-r-bigelow:
Hmm... maybe that's what felt a little funny to me. The plot is named in terms of the data it shows, not what it looks like (most of the existing components feel like they're defined the other way). That's totally cool—it's the whole "start with a dataset" vs "start with a visualization" thing. I imagine a plot like this might feature prominently in the "Example projects" section (with data pre-wired), but the visualization itself might be buried in a "Clinical" category in the "Choose a visualization" interface.

I guess this means that eventually Resonant Lab should be responsible for hard-coding which candela components are general-purpose, not candela components announcing their general-purpose-ness on their own.

PR #311, opened 2016-07-21
Survival Plot test

ronichoudhury:
I had forgotten to add an image test for the new `SurvivalPlot` component (#310)


jeffbaumes:
LGTMBTINH. BulletChart seems to have an issue - should we merge and hope this goes away?

ronichoudhury:
`BulletChart` has been exhibiting this problem for a little while now. I'm going to fix it independently but I think since the rest of the tests pass this is ok.

PR #317, opened 2016-07-26
Mapreduce cleanup

alex-r-bigelow:
Fix minor problems resulting from how the mapreduce code is stitched together


ronichoudhury:
LGTM

PR #322, opened 2016-07-27
Only make use of extraParameters when it's a JSON string

alex-r-bigelow:
database_assetstore sends non-JSON `"limit=1"` as `extraParameters` while importing database collections (this was causing `populateGirder.py` to crash). Any `extraParameters` payload is perfectly valid; our adapter should simply revert to default behavior when it doesn't recognize it.

### DIFF
--- a/app/resonant-laboratory/server/semantic_assetstore_adapter.py
+++ b/app/resonant-laboratory/server/semantic_assetstore_adapter.py
@@ -120,26 +120,34 @@ def semantic_access(Cls, offset_limit=True):
                          contentDisposition=None, extraParameters=None, **kwargs):
             dataOffset = limit = 0
 
+            compatibleParameters = None
+
+            # Can we make use of extraParameters?
             if extraParameters is not None:
-                extraParameters = json.loads(extraParameters)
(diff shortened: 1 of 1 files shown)


ronichoudhury:
I disagree just a little bit: if we send a non-JSON string by mistake, this code path is supposed to raise an error to let us know. On the other hand, `database_assetstore` accepts both JSON extraParamters and a urlencoded variety. Is there some way we can shift the use case that's breaking this code path to simply use JSON instead? That would be a smaller fix that would maintain `semantic_assetstore_adapter`'s error reporting capabilities.

alex-r-bigelow:
No, there really isn't anywhere else we can fix this problem—our `populateGirder` script needs to hit `database_assetstore`'s `import` endpoint, and that in turn issues a call to `download` (I think to probe table/collection headers?) with the non-JSON `extraParameters`. So we'd have to make `database_assetstore` conform to the JSON-only `extraParameters` to shift this case.

ronichoudhury:
If we ask nicely, @manthey may be able to help us out ;)

alex-r-bigelow:
Closing because @manthey is awesome.

PR #333, opened 2016-08-12
Fix incorrect tooltip in the Matching interface

### LINKED ISSUE #261: Existing link's tooltip says "click to connect" when hovering
It should read "click to disconnect".

The message is correct for links that are there because a project was loaded from disk. If you create a new link and then hover over it, you will get the wrong message.

alex-r-bigelow:
Fixes #261


ronichoudhury:
LGTM

PR #334, opened 2016-08-12
Make error dialog behavior consistent between OK button and close box (return to starting screen)

### LINKED ISSUE #300: Closing an error via the close box leaves the browser in a corrupt state
Clicking OK returns the user to the starting screen; the close box should do the same thing

alex-r-bigelow:
Fix #300

Also display string errors in the error dialog box


ronichoudhury:
LGTM

Issue #280, opened 2016-07-06
Migrate to D3 v4


alex-r-bigelow:
Resonant Lab makes extensive use of D3 v3.—these bits of code should be updated appropriately now that 4.0 is official.

ronichoudhury:
:+1: but be careful as you go about this. Some kind of clientside testing strategy would go a long way towards minimizing potential pain during this process.

PR #338, opened 2016-08-13
Float parse fix

### LINKED ISSUE #331: Histogram computations parse some floats incorrectly
Example:
Load gapminder.json and look at the population histogram—lots of values are getting treated as two-digit numbers (likely that somewhere values are getting converted to 13e7, and then `parseFloat` is only pulling out the 13).

alex-r-bigelow:
Fixes #331 

Turns out the bins and counts were actually accurate; the human-readable labels were incorrectly formatted. This PR fixes these format issues, and polishes them a little (using D3.format to get nicer number labels, and fixing a couple date label oddities).

### DIFF
--- a/app/resonant-laboratory/models/Dataset.js
+++ b/app/resonant-laboratory/models/Dataset.js
@@ -1,3 +1,4 @@
+import d3 from 'd3';
 import MetadataItem from './MetadataItem';
 import { parseToAst } from '../querylang';
 import RangeSet from '../shims/rangeSet.js';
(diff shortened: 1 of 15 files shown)


app/resonant-laboratory/models/Dataset.js
```diff
@@ -1,3 +1,4 @@
+import d3 from 'd3';
```
ronichoudhury: Small style concern - the "modern" d3 way to do this is something like `import format from 'd3-format';` and then use `format()` directly in the code. What do you think?

alex-r-bigelow: Yeah, that's D3 v4... It will take a lot of revision across the app, so we tucked it all in #280. For now, I'm sticking with D3 v3.

ronichoudhury: Sounds good!

ronichoudhury:
LGTM.

PR #290, opened 2016-07-12
Filter makeover

alex-r-bigelow:
Convert the dataset interface to show histogram previews of the dataset, as well as a table view.

PR #340, opened 2016-08-15
Fix problem with only updating ghost paths between elements in the matching view

alex-r-bigelow:
This was an unintended side-effect of b237fce63c2da6b0f50d97ecfd2df0c732ba4eb8


alex-r-bigelow:
@jeffbaumes, let me know if this fixes the behavior you were seeing

jeffbaumes:
Not sure why this made a Travis build fail, but LGTM, it fixes the issue I was seeing.

alex-r-bigelow:
I think it's an old image testing issue that's cropped up periodically—I think we first saw it sometime around the time #290 was merged.

Issue #305, opened 2016-07-19
Validate that routing is relative


alex-r-bigelow:
Make sure that changing root around (e.g. multiple girder plugins, new domain name, etc) doesn't interfere with resonant lab's routing.

PR #351, opened 2016-08-16
[WIP] New girder route

alex-r-bigelow:
This preps Resonant Lab for when https://github.com/girder/girder/pull/1484 is merged

TODO:
- [ ] Still need to [set the API path](https://github.com/Kitware/candela/blob/new-girder-route/app/resonant-laboratory/mainPage.js#L23) in a more robust way

### DIFF
--- a/app/resonant-laboratory/mainPage.js
+++ b/app/resonant-laboratory/mainPage.js
@@ -20,7 +20,7 @@ import './stylesheets/girderPatches.scss';
 
 // The API root is different
 let girder = window.girder;
-girder.apiRoot = 'api/v1';
+girder.apiRoot = '../api/v1';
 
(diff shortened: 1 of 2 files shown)


alex-r-bigelow:
@danlamanna, BTW I did a force push so you'll want to update your branch. Thanks for your help!

danlamanna:
For reference, this is how I approached it with Flow which was using a dynamically generated mako template: https://github.com/Kitware/flow/compare/girder-routetable

It might be desirable to just "copy" the `apiRoot` reference and attach it to the Resonant Lab root so `api/v1` will always exist regardless of where Resonant Lab is mounted.

alex-r-bigelow:
This PR should address #305

PR #352, opened 2016-08-16
Allow command line password in populatGirder.py

ronichoudhury:
This will be mainly useful for Ansible provisioning.


alex-r-bigelow:
LGTM

Issue #353, opened 2016-08-16
Data disappears sometimes


ronichoudhury:
To reproduce:

1. Start at the splash screen.
2. Select "start with a dataset".
3. Open the library menu and select the cars data set.
4. Observe that, after a delay, the dataset appears in the panel.
5. Click on the "select visualization" button in the visualization panel.
6. Select "ScatterPlot".
7. Observe that the data in the data panel suddenly disappears.

On occasion, after step 3 you may also observe that the dataset panel doesn't open, and that the data simply doesn't appear there (or, it does appear and then almost immediately disappears).

PR #356, opened 2016-08-16
Logged out fixes

### LINKED ISSUE #346: Interface does not work when user is logged out
- [x] Fix crash on logout
- [x] Fix getHistogram access issues
- [x] Fix empty histograms
- [x] Fix ERR_INCOMPLETE_CHUNKED_ENCODING problem when getting the page from mongodb
- [ ] <del>Adopt datasets when logging in</del>

alex-r-bigelow:
Fixes UI oddities when users are logged out.

Fixes #346

### DIFF
--- a/app/resonant-laboratory/general_purpose/histogram_map.js
+++ b/app/resonant-laboratory/general_purpose/histogram_map.js
@@ -2,7 +2,7 @@
 
 counter += 1;
 if (counter >= params.offset &&
-    (params.limit === 0 || counter < params.offset + params.limit)) {
+    (!(params.limit) || counter < params.offset + params.limit)) {
   var dataRow = this;
   emit('__passedFilters__', {
     histogram: [{
--- a/app/resonant-laboratory/models/Dataset.js
+++ b/app/resonant-laboratory/models/Dataset.js
@@ -46,6 +46,9 @@ class DatasetCache {
   set cachedPromises (value) {
     this._cachedPromises = value;
   }
+  clear () {
+    this.cachedPromises = {};
+  }
   get filter () {
     if (!this._filter) {
       this._filter = {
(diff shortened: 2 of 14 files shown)


ronichoudhury:
Doesn't seem to be working :(

When I fire it up, I see (as expected) a 401 error in the console (since I'm not logged in).

I do "create new project", which seems to work just fine. Then I select a data set to load (I tried both cars and gapminder) which results in an error. I get three messages in the console, all of them 401s to `POST /item/<id>/dataset/getHistograms`.

alex-r-bigelow:
Thanks for the catch! It should be fixed now for both mongo datasets and flat files.

ronichoudhury:
Still seeing a problem. Now if I start from scratch, select "start with a dataset", then select, say, cars, the data set appears to load, then immediately disappears, and I get these errors in the console:

````
jquery.js:9203 GET http://localhost:8080/api/v1/user/authentication 401 (Unauthorized)
d3.js:670 Error: <rect> attribute width: A negative value is not valid. ("-64")
d3.js:663 Error: <svg> attribute width: A negative value is not valid. ("-32")
````

(The first error is expected, since I'm not logged in.)

alex-r-bigelow:
I know I've seen those errors while logged in, so I'm fairly confident that they're symptoms of a separate issue (relevant to #353?), but I've had trouble replicating them.

Are you getting this behavior with all the datasets, or just cars?

ronichoudhury:
Seems to be all of them. One time it worked. A few other times, I get a fatal error message about "couldn't update the reference to a copy of the <name> dataset".

If you want you can swing by my desk and I'll give you a live demo of what I'm seeing.

ronichoudhury:
Looks like the problems this PR was designed to fix, are fixed. LGTM!

PR #366, opened 2016-08-22
Add pycandela Python package

jeffbaumes:


ronichoudhury:
What should I do to test this?

jeffbaumes:
From a clean environment:

```
pip install pycandela
jupyter-notebook
```

From in the notebook, put this in a cell:
```
import pycandela
pycandela.init()
```
Should say it was loaded successfully. Then enter this in another cell:

```
pycandela.component('ScatterPlot', {'data': [{'x':1,'y':1},{'x':2,'y':2}], 'x': 'x', 'y': 'y'})
```
Should show a simple plot.

ronichoudhury:
Fantastic, it works beautifully.

Is there a reason this has to go in a `python` top-level directory (as opposed to where it was in the `app` directory)?

jeffbaumes:
I didn't feel a pip package fit in an `app` directory since it is a library.

ronichoudhury:
I think of the `app` directory as anything that makes use of Candela as a library (including, in principle, other libraries). Maybe it's named poorly. A pip package may sort of make sense to stand on its own where it is.

LGTM!

PR #376, opened 2016-08-26
Server side cleaning

alex-r-bigelow:
This PR includes features and bug fixes that were added during the port to MSKCC; I'm back-porting them here for consistency between the codebases.

### DIFF
--- a/app/resonant-laboratory/general_purpose/binUtils.js
+++ b/app/resonant-laboratory/general_purpose/binUtils.js
@@ -19,7 +19,11 @@ function coerceValue (value, coerceToType) {
   } else if (coerceToType === 'number') {
     value = parseFloat(value);
   } else if (coerceToType === 'string') {
-    value = String(value);
+    if (value && value.str) {
+      value = value.str;
+    } else {
+      value = String(value);
+    }
   } else if (coerceToType === 'date') {
     if (typeof value === 'number' && value < 3000 && value > 999) {
       // Semi-smart coercion; values between 999 and 3000 are likely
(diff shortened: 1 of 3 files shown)


app/resonant-laboratory/general_purpose/binUtils.js
```diff
@@ -19,7 +19,11 @@ function coerceValue (value, coerceToType) {
   } else if (coerceToType === 'number') {
     value = parseFloat(value);
   } else if (coerceToType === 'string') {
-    value = String(value);
+    if (value && value.str) {
+      value = value.str;
```
ronichoudhury: Just curious - what objects have a `str` field like this? I thought `ObjectId`s got turned into JSON with a specially structured `_id` field?

ronichoudhury:
The code LGTM (though I don't really know how to test it; it looks like mainly code simplification).

app/resonant-laboratory/general_purpose/binUtils.js
alex-r-bigelow: Not as far as `String()` is concerned... instead, we get "ObjectId(4a58dacf6d5fdc ... )" instead of JSON or the hash. This tweak just helps us avoid the "ObjectId(" decoration.

I guess `JSON.stringify()` might make more sense than `String()`? But if we try to coerce big, complicated objects, it's probably still best to just bin them under `[Object object]`...

alex-r-bigelow:
It's sort of indirectly tested via MSKCC (the code is identical to what's running over there), so I guess that's good enough—I'll go ahead and merge.

app/resonant-laboratory/general_purpose/binUtils.js
ronichoudhury: No, this solution is perfectly good. I was just curious where that `str` field comes from.

Issue #258, opened 2016-06-06
Create a mixin for "updatable" visualization component


ronichoudhury:
This would mainly consist of an `update(options)` method.

The semantics of the method could vary (i.e. we won't, and really can't, enforce them generally) but generally speaking, any visualization that doesn't have an `update()` method could gain one by taking all the vis prep done in the constructor, moving it into `update()`, then calling `this.update(options)` in the constructor in lieu of doing the prep directly. If updating a visualization should cause an immediate rerender, then `update()` can call `render()` directly. Other options would be to emit a signal or to do nothing.

@alex-r-bigelow, what do you think?

ronichoudhury:
I could also be convinced that this should just be part of `VisComponent`, perhaps with a default empty implementation (i.e., not one that throws an error, etc.).

alex-r-bigelow:
Yeah, we're already doing the wipe-and-re-render approach as the default behavior in Resonant Lab if an `update()` function doesn't exist—moving that logic into `VisComponent` makes a lot of sense to me.

As an aside, an `update()` feature—especially allowing a view to update **without** a full re-render—is absolutely critical if we want to enable any kind of Hanpuku-style vis customization. So my vote would be to keep it simple, where the default is to wipe and re-render if no `update()` feature exists, but let the component take over whether and how to call `render()` if it implements `update()`

PR #378, opened 2016-08-26
UI fixes

alex-r-bigelow:
This PR fixes various lingering interface problems, including:
- Minor style adjustments
- Fixing label formatting issues
- A problem with the page limit/offset parameters in histogram calculations


ronichoudhury:
@alex-r-bigelow what should I do to test?

alex-r-bigelow:
The cars dataset is probably a good example—integer labels on the x and y axis had wonky formatting (e.g. 4.00), and they shouldn't anymore.

The cars page histograms also weren't updating—if you switch pages, the histograms should now change.

alex-r-bigelow:
Oh, and I guess make sure that nothing looks weird when you open tabs (you should see a nice loading gif instead of interface artifacts)

ronichoudhury:
I noticed that when I page the dataset, the histograms do recompute. However, the entire panel is now replaced by a blank view with the spinner gif. The spinner is really great, but I don't think the whole panel should disappear like that - it's disorienting because it destroys any sense of continuity you have between the old and new view, and it makes you think (however momentarily) that something has gone wrong.

I'm not sure if this PR introduced that change, but I won't stop it from merging this. I will plan on taking a look at how this works, and maybe just put the spinner as an overlay (or perhaps in the title bar) so as to preserve that sense of continuity as things change.

ronichoudhury:
I also see the spinner in the vis panel when I open it, but it doesn't seem consistent. When I create, say, a scatterplot, then I close and open the vis panel, I sometimes see the spinner and sometimes don't, but the old visualization is always there.

What governs whether the spinner appears? And what should we do about the "old" visualization (which may not actually be old) being there in a frozen, half-formed state? I suspect once we figure out the "update" protocol for Candela components, most of this may be solved, but I'm not sure if having the spinner there is helpful (I might be wrong).

ronichoudhury:
(The other stuff discussed in this PR - tick marks and zeros, fixed histograms, etc. - all look good.)

alex-r-bigelow:
Sorry about the late replies.

I set the spinner to display with every main render call, and then hide it whenever the panel is ready. In DatasetView, we show the spinner + hide everything else first, and then when the promise resolves, hide the spinner and show the UI elements. It shouldn't be too tricky to keep the other elements visible.
However, with the VisualizationView, we hide the candela component + show the spinner, and only swap after both the data is available to the visualization, and the visualization has successfully rendered.  It should be possible to do this more elegantly, but that seemed to make the most sense at the moment (definitely worth revisiting when #258 is resolved).

In both cases, we stupidly *always* show the spinner, and hide it after a promise resolves (where it resolves immediately, then you might not see the spinner).

alex-r-bigelow:
I'll remove the WIP once I can get the logo fixes in... or if we'd prefer to get the other fixes in quickly (particularly the fix for #345), we can merge now and fix the logo stuff on a different PR.

ronichoudhury:
What logo fix do you mean?

alex-r-bigelow:
Mainly, don't hide the whole dataset panel while the spinner is showing (probably grey out what was in the panel, and show the spinner on top).

I was also thinking I should try to make the issues you point out in the vis panel a little more consistent / less jarring (probably just do the same grey-out thing as the dataset panel).

ronichoudhury:
Let's not add new features to this PR. We can make new issues from the stuff I pointed out. If you're done pushing commits for now, I'd like to QA this one last time before merging. Is that an ok plan?

alex-r-bigelow:
Sounds great. I think the things to watch for are:
- Do the page histograms render correctly (with / without filters)?
- Do the labels (x and y) in the histograms make sense (e.g. no unnecessary zeros)?
- Are the histogram calculation oddities fixed for both:
   - Loading a dataset via the UI (does everything work while loading a big-ish dataset)?
   - Loading a dataset via the setup script doesn't crash when there aren't any mongo datasets (e.g. try `populateGirder.py -t 10000000000000` to force all datasets to be loaded as flat files)

ronichoudhury:
Ah, thanks for the checklist!

I'll try to get to this later today (but no promises).

alex-r-bigelow:
Hmm... the bar/grid view showing up is a bug, not a feature—somehow I must have inadvertently turned off hiding it. The switch should behave as you expect. I can try to fix this in a day or two.

ronichoudhury:
I poked around some, and I am stumped. It looks like a problem with the JQuery show/hide calls, but I can't diagnose it. The show/hide calls seem to have no effect, but from the console if I make those calls manually they work beautifully.

I'll take another look tomorrow but Alex if you get to it I'll be curious what you find.

ronichoudhury:
There was a line at the end of `render()` that made the overview, histogram, and table all visible for some reason. Removing that line seems to have fixed the problem.

ronichoudhury:
Since I'm on the hook for approving these changes but I made some changes myself, could someone give me an LGTM before I re-approve this PR?

app/resonant-laboratory/views/widgets/DatasetView/template.html
```diff
@@ -1,5 +1,5 @@
-<img id="noDatasetState" class="emptyStateHelper" src="../../../images/emptyState/noDataset.svg"/>
-<div id="datasetOverview">
+<img id="noDatasetState"  style="display: none;" class="emptyStateHelper" src="../../../images/emptyState/noDataset.svg"/>
```
alex-r-bigelow: I tend to dislike doing this in the template if there's already logic in the javascript to show/hide pieces. But that's a personal style preference, so this is fine for now.

ronichoudhury: I agree, but what's the alternative? The problem is that a bunch of jumbled stuff shows up as soon as the template is instantiated, then disappears when the appropriate show()/hide() calls go occur.

PR #380, opened 2016-08-31
Fix inferAll to allow fields with embedded periods

jeffbaumes:
`datalib.type.all` and `datalib.read` were failing when datasets had field names with embedded periods. This implements and tests a new utility function `inferAll` which calls `datalib.infer` in a way that ensures periods won't be split.

### DIFF
--- a/src/candela/util/test/util.js
+++ b/src/candela/util/test/util.js
@@ -1,0 +1,9 @@
+import test from 'tape-catch';
+import { inferAll } from '..';
+
+test('inferAll()', t => {
+  t.deepEqual(inferAll([{'a.b': 1}]), {'a.b': 'integer'}, 'should work on fields with nested dots');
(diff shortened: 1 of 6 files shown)


src/candela/util/test/util.js
```diff
@@ -0,0 +1,9 @@
+import test from 'tape-catch';
+import { inferAll } from '..';
+
+test('inferAll()', t => {
+  t.deepEqual(inferAll([{'a.b': 1}]), {'a.b': 'integer'}, 'should work on fields with nested dots');
+  t.deepEqual(inferAll([{'a': {'b': 1}}]), {'a': 'string'}, 'should only accept top-level fields');
+
```
ronichoudhury: Could we maybe add some more tests to show that this behaves the way the old function worked that it's replacing?

jeffbaumes: Done.

ronichoudhury: Nice test!

ronichoudhury:
LGTM

PR #381, opened 2016-09-01
Initial R htmlwidgets integration

jeffbaumes:


jeffbaumes:
It appears the error here is the bulletchart image test, which this PR should not affect.

ronichoudhury:
This LGTM, going by the demo I got last week. If you think it's in good shape, let's merge it.

PR #382, opened 2016-09-08
Update R docs for non-development install

jeffbaumes:
Now that it is on master, it can be installed in R directly from GitHub, so the docs now highlight that install procedure.


ronichoudhury:
LGTM

PR #384, opened 2016-09-09
Use inspect module to determine plugin base directory

ronichoudhury:
This PR replaces a brittle mechanism (using the `subprocess` module to ask `girder-install` what the Girder plugin path is) with a more robust one that uses the `inspect` module's reflection capabilities to figure out where the calling module, and therefore the `resonant-laboratory` directory, is.

### DIFF
--- a/app/resonant-laboratory/server/pluginDir.py
+++ b/app/resonant-laboratory/server/pluginDir.py
@@ -3,6 +3,12 @@ import os
 
 
 def pluginDir():
+    # This function works by grabbing the current stack, examining the first
+    # frame (which refers to the invocation of this very function), and
+    # extracting the second element of it, which refers to the filename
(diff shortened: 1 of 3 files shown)


app/resonant-laboratory/server/pluginDir.py
```diff
@@ -0,0 +1,10 @@
+import inspect
+import os
+
+
+def pluginDir():
+    me = inspect.stack()[0][1]
```
jeffbaumes: A comment might be warranted on this line to explain what it does.

jeffbaumes:
LGTM other than suggestion to add a comment

ronichoudhury:
@jeffbaumes can you check that comment to see if it's sufficient?

jeffbaumes:
LGTM, thanks.

PR #389, opened 2016-09-19
Update Ansible/Vagrant process

ronichoudhury:
This makes Vagrant/Ansible provisioning work, providing some options to the provisioning process for setting certain values in the deployment, such as the admin user password, and which git revisions to use for dependent software, etc.


app/resonant-laboratory/ansible/xenial/girder.service.template
```diff
@@ -0,0 +1,7 @@
+[Unit]
+Description=Girder
+After=network.target
+
+[Service]
```
danlamanna: At a minimum I would put a User/Group here so Girder doesn't run as root. Example [here](https://github.com/girder/girder/blob/master/devops/ansible/roles/girder/templates/daemon/girder.service.j2#L6-L7). Same goes for the upstart config.

app/resonant-laboratory/ansible/library/girder.py
```diff
@@ -1,5 +1,5 @@
 #!/usr/bin/python
```
danlamanna: If you wanted, you could use the galaxy role for no other purpose than to bring this module in automatically.

app/resonant-laboratory/ansible/xenial/girder.service.template
ronichoudhury: I'll need to confer with IT before I make this change, so here's an issue describing the problem: #390

app/resonant-laboratory/ansible/library/girder.py
ronichoudhury: Cool. I'll be in touch about this.

PR #392, opened 2016-09-19
Install bin bounds for ordinal bins

ronichoudhury:
Without `lowBound` and `highBound` properties on the returned bins, the client will think these are categorical bins.


ronichoudhury:
@alex-r-bigelow PTAL

ronichoudhury:
Thanks Alex!

jeffbaumes:
Thanks for tracking this down @ronichoudhury

PR #394, opened 2016-09-20
Use a better default colormap for Vega components

ronichoudhury:
Instead of a near-white-to-green colormap, this institutes a red-to-blue colormap that is familiar to viewers and will show up better in print.

This PR also attempts to fix the longstanding problem with the bulletchart image test.


ronichoudhury:
All tests passed! Woo hoo!

ronichoudhury:
Thanks @jeffbaumes. I'll file an issue about selecting softer versions of these colors.

PR #403, opened 2016-09-21
Replace complex logo with simpler Lora logo

ronichoudhury:
- Changes logo/name to Lora version with logo:
  - colors appear on rollover in navbar
  - white watermark brightens on rollover in app background
- Changes body typeface from Josefin Sans to Source Sans Pro


ronichoudhury:
We need to choose between this PR, #404, or a hybrid of the two. Here is a summary:
- this PR includes white logos that highlight as you mouse over; #404 has the same, but they colorize when you mouse over. The logos appear in the upper left (inside the nav bar) and as a watermark on the background of the app
- both PRs include a full-color logo inside the "about" page

To hybridize, we might choose to colorize one or the other of the two white logos.

If we do so, my opinion is to leave the "watermark" logo as white. The color change is such a large logo is a bit jarring. It works better for the upper left logo.

However, leaving them both white has a nice subtlety as well.

@jeffbaumes and @alex-r-bigelow, please weigh in.

alex-r-bigelow:
To be honest, I'm not a fan of Lora + Josefin Sans... it feels really weird. Would it be possible to switch to Josefin Slab? Or if we stick with Lora, I'd prefer to change the typeface across the app as well. I know that's kind of a picky thing, but these sort of typeface mismatches drive me crazy.

I'll try out the color vs white stuff when I get a chance (probably tomorrow afternoon).

ronichoudhury:
We tried Josefin Slab for the logo and it just doesn't look right. I actually think Lora goes better with Josefin Sans then Josefin Slab does :disappointed: 

In any case, Lora is the typeface used across other Resonant projects, so I think we have to center around that. I will try a variant with Lora in the app instead of Josefin to see how it looks/feels.

ronichoudhury:
Lora for the rest of the text doesn't seem to work too well. A sans serif typeface seems like it would fit better.

According to Google Fonts, Open Sand, Lato, Raleway, Montserrat, and Roboto are popular accompaniments for Lora.

Out of those, Open Sans is inoffensive; Roboto is much like Open Sand but more compact, and Raleway has a similar feel as Josefin, but more "orthonormal".

I'll create a pair of PRs using Roboto and Raleway for comparison (both rooted on this PR, to prevent a combinatorial explosion).

alex-r-bigelow:
Thanks for humoring me. Though I think the designer of Josefin (sans / slab) might disagree with you... :suspect:  

I'll check out the other font combos when I get a chance.

alex-r-bigelow:
After further review, I guess if we're going for the resonant.kitware.com feel, we definitely should ditch Josefin and go with Montserrat+Roboto Slab+Lora (that's what it uses)... I don't think Raleway worked too well, and the more we can avoid adding typefaces to the party, the better.

I'll start a quick branch with those in place (Montserrat for the all-caps headers, Roboto for the body text.) Note that we also need to update the empty state text (it's an SVG like the logo).

As far as the white/color issue goes, I like your idea of just using color on the header, but keeping the watermark white. I'll add these changes in the branch as well.

ronichoudhury:
Sounds good, Alex. Let's check out that combination and see how we like it. Feel free to open a WIP pull request on that branch when the changes are in place.

ronichoudhury:
Did a review of the app and a few typeface choices with Steve today. He suggested picking just two typefaces - one for the logo/headings, and one for text. Against black or dark backgrounds, he suggested thin, white fonts (such as in the window title bars). And generally we looked for more condensed fonts so as to keep blocks of text, and labels for icons (e.g. "Create an Empty Project", etc.) trim and elegant.

After touring a few choices, we found Source Sans Pro, which has somewhat more character than Open Sans / Helvetica / et al. but still works as a no-nonsense, easy to read choice.

I realize this is not the most exciting typeface choice, but we do need to nail something down and move on. When we look over style choices across the Resonant suite, we can certainly revisit the typeface choices for Resonant Lab as well.

Unless there are any strenuous objections, I'd like to get this merged. Thanks for discussion and consideration!

Issue #400, opened 2016-09-21
Histogram selections do not propagate to the visualization


jeffbaumes:
The visualization is visibly refreshed, but the data does not change until the page is refreshed.

ronichoudhury:
Closed via #407

PR #407, opened 2016-09-22
Remove faulty caching mechanism for updated data requests

ronichoudhury:
After filters are changed, the histograms and data panel view of the new data update properly, but the visualization never receives the update, instead relying on a out-of-date, cached promise containing data from the first download thereof.

This simply removes the caching mechanism, allowing the visualization to update properly.

@alex-r-bigelow, I am not sure what the error in logic was; maybe you can diagnose that.

@jeffbaumes if you test and approve this, I can push the fix up to our deployment instance.


ronichoudhury:
Fixes #400

alex-r-bigelow:
I dug into this a little; I think part of the problem should be fixed by #378. The remaining bit should just have something to do with the cache not getting cleared when the filters are changed... I'll take another look sometime next week if I get a chance.

PR #408, opened 2016-09-22
[WIP] Match resonant website

alex-r-bigelow:
See conversation in #403


ronichoudhury:
I don't think Roboto Slab is right for the app, though it may just be because I am used to how it was before. The slab font is too heavy/bulky in, e.g., the matching panel. I think Roboto or even Roboto Condensed might work better (but of course I'd have to see it).

I may grab Steve Jordan tomorrow and get his opinion as well.

alex-r-bigelow:
Do you mean the matching panel header, or the clickable pills inside? I thought the latter was one of the few places it *wasn't* too heavy. If you meant the header, that's actually Montserrat.

Both fonts have lighter weights available—If we want to match the Resonant website, we should use the same fonts as much as possible (changing the weight is usually a far less drastic change than dropping the slabs / going condensed).

ronichoudhury:
Right, I meant inside the pills. I like the Montserrat caps in the window titles.

I actually don't think we need to make it an explicit goal to match the fonts in the Resonant website. If we do, that's ok, but the fonts need to serve their purpose within the app first and foremost.

jeffbaumes:
Another option that I'll throw out there is that we can change the fonts on the Resonant website :) The main Girder client and various RTD are two others areas that would benefit from style normalization.

ronichoudhury:
I'll take the bold position that our attempts to figure out typefaces for
Resonant Lab should not become a larger discussion about fonts on our other
websites at the current time :-P

alex-r-bigelow:
In that light, it might be worth waiting for Kitware to implement some kind of branding style guide (I think @superlib mentioned something about revamping the main page—maybe all sites should conform to whatever styles are put in place there?)

jeffbaumes:
Sorry for further devolving the conversation :) I agree with @ronichoudhury: make the app look good - if there are matches with the current Resonant website that's great but not required. I just wanted to add that the Resonant website did not have an in-depth style discussion so should not be considered gold-standard.

I like the energy around making RL look awesome, but for now let's pick a "good enough" and  we can always continue iterating later.

ronichoudhury:
I'm planning to sit down with Steve later today, so we'll try to figure out
a combo that works generally. When overarching style guides come in some
day, we can reconsider RL as well.

ronichoudhury:
Closing in favor of #403

PR #416, opened 2016-09-28
Renames Resonant Laboratory to Resonant Lab

ronichoudhury:
This will break any existing Girder plugin you have set up for Resonant Lab. To avoid problems, first go into your running Girder's admin console and deactivate the Resonant Lab plugin. Make sure your symlink/directory in the plugins directory is called "resonantlab". And rename the public library collection in Girder to "Resonant Lab Library".

I have also tested the Vagrant/Ansible infrastructure and it seems to be working properly.

This PR is also likely to conflict with any work you have in flight. Apologies in advance :)


ronichoudhury:
Does this look good otherwise?

jeffbaumes:
I've fixed up a few more instances so now `git grep laboratory` and `git grep Laboratory` are clean.

ronichoudhury:
Right, so I think generally if someone besides the PR author makes commits, someone besides the committer should manually "LGTM" the changes (which I failed to do in my "thanks" comment, even though that's what I meant) in addition to someone providing the official PR approval.

PR #417, opened 2016-09-28
Fix date handling by using read() instead of inferAll() in axis logic

### LINKED ISSUE #409: Dates not compatible between Resonant Lab and Candela Vega components
I am able to get Candela's Vega plots to take string dates like `"2014-01-01"` and produce valid date axes, but from Resonant Lab I cannot get the dates to transfer correctly. This may be because the Vega charts don't respond well to actual Date objects, while the RL data panel histograms expect them to be Date objects.

I expect the best fix will be to make the Candela Vega plots resilient to accepting Date objects.

jeffbaumes:
Fixes #409.

It turns out that not handling dates was a bug introduced by #380. We needed to be more careful about keeping the logic of datalib's `read()` instead of using `inferAll()` (see change in src/candela/util/vega/index.js). `read()` does the proper conversions of dates (including Date() objects) into the normalized millisecond format for use by Vega scales/axes.

This branch adds testing that more directly compares our tweaked versions of `inferAll()` and `read()`, making them function identically to the datalib versions except in the case of having periods in the field names, which our versions handle.

If my PR to datalib (or something like it) is accepted and released, we will not need our own versions of `inferAll` and `read`. See https://github.com/vega/datalib/pull/80.

### DIFF
--- a/src/candela/components/UpSet/index.js
+++ b/src/candela/components/UpSet/index.js
@@ -1,7 +1,7 @@
 import d3 from 'd3';
 import { unique } from 'datalib';
 import VisComponent from '../../VisComponent';
-import { inferAll } from '../../util';
+import { read } from '../../util';
 import * as upset from 'UpSet';
 import template from './template.html';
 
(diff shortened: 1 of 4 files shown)


ronichoudhury:
@jeffbaumes are you waiting on anything else to merge this PR?

jeffbaumes:
No, just somehow missed that you approved it. Thanks.

ronichoudhury:
Woo hoo!

PR #418, opened 2016-09-29
Minor style refinements

alex-r-bigelow:
This PR includes a few minor tweaks to the appearance of the app to conform to recent style changes (I'm prepping the VIP slides / poster, and my OCD is kicking in):
- Switch the empty state images to use Source Sans Pro instead of Josefin
- Dial the bold variant of Source Sans Pro back a little bit (900 -> 700)


jeffbaumes:
Thanks, you're not the only one who can be a ✨perfectionist✨. Those two things were nagging at me as well.

PR #421, opened 2016-10-03
Fix export of SVG by passing format to serialize

### LINKED ISSUE #399: Export SVG produces same result as Export PNG
To reproduce: make a scatter plot, then export both as PNG and SVG, and then run `cmp` on both files.

jeffbaumes:
Fixes #399 

Silly omission.


ronichoudhury:
Does `serialize()` do PNG export by default? I kind of think it should have to be explicit, even if PNG output is the (presumed) common case. What do you think @jeffbaumes?

PR #424, opened 2016-10-04
Update onset to prevent CSS leakage

### LINKED ISSUE #402: Prevent leaking of CSS in LineUp, OnSet, UpSet
Some CSS is leaking through and is not namespaced within the various components.

jeffbaumes:
Fixes #402 

normalize.css was being pulled in through OnSet, which does funny things to global elements when included after other CSS, such as adding padding around `ul` tags. This was resolved in https://github.com/Kitware/setvis/pull/3, and this PR updated OnSet to the new version.

There is still special CSS logic in the sphinx docs to deal with this, which we need to keep until this version of Candela is on CDN - another reason to move forward with semantic-release / commitizen.


ronichoudhury:
Do you mind opening a new issue about removing that special logic once we have a CDN ecosystem spun up?

PR #428, opened 2016-10-06
Enable semantic release

ronichoudhury:


ronichoudhury:
Here we go... :point_right: :red_circle:

PR #433, opened 2016-10-07
Add Google analytics infrastructure to Resonant Lab build

ronichoudhury:


danlamanna:
I don't really know the scope of this, but is this something the Girder [Google Analytics](http://girder.readthedocs.io/en/latest/plugins.html#google-analytics) plugin can be used for?

jeffbaumes:
@danlamanna Don't think so because that plugin is for the main Girder interface while this is an entirely separate UI.

ronichoudhury:
The main thing here is to confirm with Sandy that analytics data is indeed coming in. I'm going to check with her in a few days.

I will send you some info on testing Ansible deployments; if that info needs to go into the docs we can do that as well.

PR #434, opened 2016-10-07
Validate commit messages using commitizen and git commit hooks

ronichoudhury:
This may be slightly controversial. Even if the simple implementation in this PR is insufficient, I'd still like to figure out a way to have this type of enforcement at the commit message level.

The most latitude we have here is in two, or maybe three, places:
- the `helpMessage` string in [package.json line 138](https://github.com/Kitware/candela/blob/validate-commit-msg/package.json#L138) can explain gently what happened and how to fix it
- we should have a CONTRIBUTING.md that explains this in detail
- the developer documentation could discuss this as well

Probably the help message should point to CONTRIBUTING.md. The dev docs could too.

As I write this PR description, that seems to be what we should do. I'll mark this WIP until we confirm that we really need a CONTRIBUTING.md now, and it can become part of this PR.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,168 @@
+# Contributing to Candela
+
+Candela is an open-source project and therefore relies on your contributions in
+conjunction with the efforts of the core developer team to grow and improve. You
+can contribute to Candela in many ways, not all of them involving writing code.
+We recommend the following checklist of ways you can help:
+
+   * Read the [Candela documentation](http://candela.readthedocs.io/en/latest/)
+     to become familiar with usage and development of the library. If the
+     documentation is unclear or can be improved in some way, please let us
+     know.
+
+   * Ask us a question on the [Candela user mailing list](). The mailing list
+     is a good place to discuss problems you may run into that are not addressed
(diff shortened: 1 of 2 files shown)


jeffbaumes:
I think a CONTRIBUTING.md is in order on this PR, along with a pointer from the sphinx docs to CONTRIBUTING.md.

Girder's CONTRIBUTING.md is probably a good place to start, though commitizen related things would need to be added.

ronichoudhury:
Warning: I force-pushed this branch after rebasing against master.

ronichoudhury:
@jeffbaumes I added a CONTRIBUTING.md that's ready for review.

I'm leaving the WIP tag in place until we decide whether to include mailing list info in this PR (which in turn is also pending on the actual creation of mailing lists).

jeffbaumes:
The text all looks good. If you want to merge now, I'd suggest removing the mailing list bullet point. I'd gladly approve with that one change. Or you can wait for the mailing list setup.

ronichoudhury:
Let's go ahead and merge now. I'll excise the bullet point... stand by.

ronichoudhury:
Ready for final review!

PR #440, opened 2016-10-11
chore(resonantlab): Point to candela docs from vis selector

### LINKED ISSUE #247: Point to candela docs from Resonant Lab interface
Somewhere for each vis (perhaps from the maching interface) there should be a link to the README.md markdown for that vis that opens in a new tab. That will help people know what the various inputs mean.

Bonus points for pulling down the markdown and displaying it in the app itself, but that is probably overkill at this point.

jeffbaumes:
Fixes #247

This is a start that is not super pretty but at least lets the user go to the docs to find out more about each vis type.

### DIFF
--- a/app/resonantlab/views/overlays/VisualizationLibrary/index.js
+++ b/app/resonantlab/views/overlays/VisualizationLibrary/index.js
@@ -46,6 +46,13 @@ let VisualizationLibrary = Backbone.View.extend({
     libraryButtons.select('span')
       .text(d => d.name);
 
+    libraryButtonsEnter.append('a');
+    libraryButtons.select('a')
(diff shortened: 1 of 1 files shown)


jeffbaumes:
Looks like we have a breaking change in Girder Client, I assume because of the 2.0 release. We can pin it to 1.x or update our script. @zachmullen can you point us to the solution (in `script/upload_test_images.py`)?

zachmullen:
Yep, I'll put in a fix.

zachmullen:
If you merge master or rebase this, it should work.

jeffbaumes:
Ok this is rebased and force pushed.

PR #461, opened 2016-10-24
Disable versioning logic to prevent orange warnings in deployments

ronichoudhury:


jeffbaumes:
This does not seem to remove the orange circles in all cases. May be safer to keep the server logic but disable the display logic in the JS code.

app/resonantlab/views/overlays/ProjectLibrary/index.js
```diff
@@ -162,7 +162,7 @@ let ProjectLibrary = ProjectSettings.extend({
           if (d.attributes.meta.rlab.versionNumber === appVersion) {
             return 'none';
           } else {
-            return 'block';
+            return 'none';
```
jeffbaumes: DataSetLibrary/index.js has a similar need for a change.

ronichoudhury:
Should I revert the change to the server logic, now that displaying the warning itself is suppressed?

jeffbaumes:
Yeah, I'd back out the server logic change.

ronichoudhury:
@jeffbaumes PTAL

PR #462, opened 2016-10-25
Accept windows excel file mimetype as CSV file

ronichoudhury:


jeffbaumes:
I think this is good for now, but I wonder if it will think it can accept .xls files, when at this point it can't.

ronichoudhury:
I was thinking about this some more and I'm not sure why we gate files by their declared mimetype or extension, when it is relatively easy to sneak a file through only to have it fail parsing later. The app doesn't currently show an error for a failed parse, so I think it might be better not to gate files at all (aside from showing a warning in place of the error message Lisa was getting), and then handle bad formatting or other parsing failure as the actual point of failure for file uploads.

In that case, Lisa's files would have "just worked", while other files would fail to parse, show an error, and then revert to the state before the dataset selection began.

PR #465, opened 2016-11-04
Add initial TrackerDash component

ronichoudhury:
attn: @cpatrick and @mgrauer


src/candela/components/TrackerDash/ErrorBulletWidget.js
```diff
@@ -0,0 +1,69 @@
+import _ from 'underscore';
+import d3 from 'd3';
+import nv from 'nvd3';
+import $ from 'jquery';
+
+import { computeColor } from './utility.js';
+import VisComponent from '../../VisComponent';
+
+class ErrorBulletWidget extends VisComponent {
```
jeffbaumes: Should these be called "Widget" or should that suffix be removed?

trackerdash/README.md
```diff
@@ -0,0 +1,47 @@
+## TrackerDash JQuery Plugin
```
jeffbaumes: This isn't a jQuery plugin anymore is it? Or is this some other add-on thing?

src/candela/components/TrackerDash/ErrorBulletWidget.js
ronichoudhury: I'm ok either way - this is still an "internal" component of TrackerDash, so if/when we promote this to a top-level component, we'd certainly want to change the name. I left the name as-is since I didn't vet any of the design for any "internal" components for general-purposeness; essentially, I'm still thinking of these as specifically designed for TrackerDash.

Let me know if you think we should change the name now, and do a heavier code-review type of step later.

trackerdash/README.md
ronichoudhury: Nope, no more jQuery. I'll change the README.

ronichoudhury: Actually, the files in this directory, besides the README, were slated for deletion. I forgot to get rid of them. I'll also transfer the README to the appropriate directory, and change it to reflect this PR.

src/candela/components/TrackerDash/ErrorBulletWidget.js
jeffbaumes: I'm ok with keeping Widget in the name for now. I agree we can change it later if needed.

ronichoudhury:
@jeffbaumes, @mgrauer: I've found a fix for the styling issue I showed both of you on this branch.

The problem was being caused by the Bootstrap and NVD3 style files used by the TrackerDash component. I guess these introduce global changes that are hard to control (since they come from published NPM packages).

My solution for now is to make the client inject these styles at component instantiation time. They could also inject other styles or no styles at all as well. This leads to a larger question of how we want the TrackerDash component stuff to shake out in the steady state. I assume we want it to be more like a TrackerDash framework, where people can construct their own dashboards from whatever components they like, including the ones we're introducing in this PR. In that case, styles would not be tied to the component per se; they'd go with whatever agglomeration of things different users need for different dashboards.

PR #473, opened 2016-12-23
Export "webpack helper" function to aid usage in downstream Webpack projects

ronichoudhury:
- Publishes a function in `candela/webpack.js` to aid in adding loaders and other peculiar-to-Candela configuration to a downstream project's Webpack configuration in a black-box way
- Moves source files from `src/candela` into the top level of the repo - this makes it easier for downstreams to include Candela sources in their projects:
````javascript
import LineChart from 'candela/components/LineChart';
````
instead of 
````javascript
import LineChart from 'candela/src/candela/components/LineChart';
````
- Updates the LineUp and UpSet dependencies to versions that don't cause problems when Candela is used downstream this way

TODO:
- [x] Generate new LineUp image test after verifying that the component is behaving properly
- [x] Adjust `.npmignore` to restrict what gets published to NPM
- [x] Update documentation

### CODE REVIEW

**Comment by ronichoudhury** [comment]
@mgrauer, we can use this branch to try out importing Candela piecemeal into TrackerDash.

**Comment by mgrauer** [comment]
Is this ready for me to try, or would it be better to debug it together?

I've got in my downstream

```
import 'nvd3/build/nv.d3.min.css';
import './main.styl';
import candela from 'candela';
  
window.TrackerDash = candela.components.TrackerDash;
```

should that change?

:snowman_with_snow:

ronichoudhury:
You should be able to try it yourself, though I assume we will need to follow up with some debugging (either together or just me). Here are the steps to try it out:

1. Change your Candela dependency in package.json to `"git+ssh://git@github.com/kitware/candela.git#webpack-helper"`.
2. Remove `node_modules/candela`.
3. Run `npm prune` if you like.
4. Run `npm install`.
5. Change the Candela import statement to

   ```javascript
   import TrackerDash from 'candela/components/TrackerDash';
   ```

   and the next line to

   ```javascript
   window.TrackerDash = TrackerDash;
   ```
6. Rebuild the project.
7. Call/email/text/telegraph Roni with the list of errors.

jeffbaumes:
Interesting. So this removes src and puts the components dir at the top level?

ronichoudhury:
Yes, and also the `VisComponents` and `util` directories (essentially, everything "top-level"). It makes for a bit of an uglier codebase, which is the cost of a smoother downstream experience.

ronichoudhury:
This is ready for review.

jeffbaumes:
When building from scratch just now, the Resize and Serialization examples give a `Uncaught SyntaxError: Unexpected token import`. Also, OnSet has an error saying `$` is not found.

jeffbaumes:
I pushed a fix for the Resize and Serialization examples. It was because candela.js was not going through Babel. I'm ok with the OnSet error getting offloaded to another issue that we solve separately.

ronichoudhury:
Thanks Jeff. I notice that in addition to the undefined-`$` problem, the resize examples don't actually resize. We can address that separately as well.

PR #479, opened 2017-04-19
Add similarity graph component

ronichoudhury:
#### TODO
- [x] Add documentation
- [x] Add options spec (to make this work provisionally in Resonant Lab)
- [x] Fix failing style tests
- [ ] ~~Investigate/fix failing image tests~~
- [x] Add size parameter

EDIT: The failing image tests might have to do with the version change on DKC, so I will investigate that in a separate PR.

@jeffbaumes PTAL

Build the library with `npm run build` and then the examples with `npm run build:examples`. Then, launch the example server with `npm run examples`, and finally go to http://localhost:3000/examples/similarity/ to see an example using bigram frequencies as an example.


ronichoudhury:
@jeffbaumes this is ready for review

docs/components/similaritygraph.rst
```diff
@@ -0,0 +1,139 @@
...
+
+    pycandela.components.BarChart(data=data, x='a', y='b')
+
+**R**
+
+.. code-block:: r
+
+    library(candela)
+
+    candela('BarChart', data=mtcars, x='mpg', y='wt', color='disp')
```
jeffbaumes: Python and R examples refer to BarChart.

package.json
```diff
@@ -69,6 +69,7 @@
     "conventional-commit-types": "^2.1.0",
     "copy-webpack-plugin": "^1.1.1",
     "cz-conventional-changelog": "^1.2.0",
+    "d3": "^3.5.17",
```
jeffbaumes: Does cola not support d3 v4? I have been using d3 v4 for some other things I'm hoping to get in Candela.

components/SimilarityGraph/index.js
```diff
@@ -0,0 +1,173 @@
...
+          fieldTypes: ['number', 'integer']
+        }
+      },
+      {
+        name: 'size',
+        type: 'number',
+        domain: {
+          mode: 'value'
+        }
+      }
```
jeffbaumes: See https://github.com/Kitware/candela/blob/tree-heatmap/components/TreeHeatmap/index.js#L65-L72 for how I specified a numeric input for the TreeHeatmap component that works with @jbeezley's UI generation.

app/examples/similarity/index.js
```diff
@@ -0,0 +1,59 @@
...
+  const rescale = val => (val - min) / (max - min) * (maxSize - minSize) + minSize;
+
+  table.forEach(entry => entry.size = rescale(entry.size));
+  console.log(table);
+
+  return table;
+};
+
+window.onload = () => {
+  showComponent(SimilarityGraph, 'svg', {
```
jeffbaumes: Can you modify this so that the containing element is a `div` instead of an `svg`? This would make candela-using logic simpler in places like resonant lab, girder candela plugin, etc.

package.json
ronichoudhury: I think it does, but I was copying patterns from an earlier project to get this working here.

Would you prefer if I upgrade to D3 v4 right in this PR, or defer to a future PR?

app/examples/similarity/index.js
ronichoudhury: I can, but doing so brings up two questions/concerns:

1. If the container type is `div`, then the component constructor will have to create its own SVG element within the container. Is that ok?

2. Related in a way - if we take the approach of constructing our own SVG, would there be any component at all that *doesn't* take a div as its containing element? If not, then we should think about dumping the "container type" specifier that currently goes with `VisComponent`s.

docs/components/similaritygraph.rst
ronichoudhury: Oops. I also see that I forgot to update the options list down below as well.

The example data I used here was quite a bit more complicated than for the other components. Should I replicate the data or just skip the Python/R examples?

components/SimilarityGraph/index.js
ronichoudhury: Take a look at 3413505 to see if I got this right.

docs/components/similaritygraph.rst
jeffbaumes: We could probably come up with a simple 3x3 similarity matrix plus an id column that would work for an example. I think I had something fairly simple in my problem description email.

package.json
jeffbaumes: I'm not sure what, if any, v4 features I am using, but I think vega uses v4, and it makes sense to use it since it's been out a while. I'd prefer just doing it now - if it is taking more than a few hours to upgrade we can consider holding off.

app/examples/similarity/index.js
jeffbaumes: > If the container type is div, then the component constructor will have to create its own SVG element within the container. Is that ok?

Yep - that is what e.g. Vega does all the time in SVG rendering mode.

> Related in a way - if we take the approach of constructing our own SVG, would there be any component at all that doesn't take a div as its containing element? If not, then we should think about dumping the "container type" specifier that currently goes with VisComponents.

The use case we were considering is if we had GeoJS layer components whose parent was a GeoJS container. I agree that without that, there is no current use case for components specifying a container type.

docs/components/similaritygraph.rst
jeffbaumes: Implicit in my response what that the Python/R examples don't need to replicate the exact graph that is embedded in the page - they can be much simpler.

ronichoudhury: Sounds good.

app/examples/similarity/index.js
ronichoudhury: See 8bdd5a016f1d8088a78f2c067536c22799f37a15

docs/components/similaritygraph.rst
ronichoudhury: See 4527ab6e5a1976f3447446f86dfd3fee03cd6d6d

jeffbaumes:
Ok @ronichoudhury my stuff is in. I also updated the docs to reflect my changes.

docs/components/similaritygraph.rst
```diff
@@ -116,7 +117,7 @@ Example
     id = c('A', 'B', 'C')
     class = c(0, 1, 1)
     A = c(1.0, 0.5, 0.3)
-    B = c(0.5, 1.0, 0.2)
+    B = c(0.5, 1.0, 0.2)the
```
ronichoudhury: Looks like a typo?

ronichoudhury:
@jeffbaumes could you change the commitizen tag on that last commit from `feat` to `refactor`? I want to reserve `feat` for introducing a full new feature on a branch, rather than incremental improvements to a feature mentioned in a different commit.

It looks like I need to remessage a commit along these same lines.

Taking a larger view, what we really need is a `dev` tag we can use for incremental commits, with the same semantics as `refactor` ("a code change that neither fixes a bug nor adds a feature") but a better name reflecting what the commit actually does.

docs/components/similaritygraph.rst
jeffbaumes: Oops will fix.

jeffbaumes: Ok fixed.

jeffbaumes:
I should be all set with your requested changes now @ronichoudhury

ronichoudhury:
Thanks @jeffbaumes. I also force-pushed my commit log message change. If this all LGTY, it LGTM too. I think you'll still need to approve the PR formally.

PR #480, opened 2017-04-24
Tree heatmap

jeffbaumes:
This adds a specialized heatmap where row and column trees can be attached to the data matrix. It has a bit more cleanup still to do.

* [x] Documentation


ronichoudhury:
Is this ready for review?

jeffbaumes:

docs/templates/layout.html
```diff
@@ -1,6 +1,8 @@
 {% extends "!layout.html" %}
 
 {%- block extrahead %}
+    <!-- Switch the script tag for the docs to work locally. -->
+    <!-- <script src="/_static/candela.js"></script> -->
```
ronichoudhury: There must be a way to do this using templating magic? We can investigate in a separate PR.

PR #481, opened 2017-04-27
Upgrade D3 to version 4

ronichoudhury:


ronichoudhury:
@jeffbaumes it turns out to be much easier than I thought - I just had to explicitly include each component's "local" copy of D3 in the component code. No need for separate wrapper packages, etc.

PR #485, opened 2017-04-29
Add SentenTree component

ronichoudhury:
sententree

Depends on #483 for proper CI.

### DIFF
--- a/components/SentenTree/index.js
+++ b/components/SentenTree/index.js
@@ -43,11 +43,18 @@ export default class SentenTree extends VisComponent {
     ];
   }
 
-  constructor (el, {data, graphs = 3}) {
+  constructor (el, {data, id = null, text = 'text', count = 'count', graphs = 3}) {
     super(el);
 
+    // Transform input data into correct form.
+    this.data = data.map((d, i) => ({
+      id: id ? d[id] : i,
+      text: d[text],
+      count: d[count] !== undefined ? d[count] : 1
+    }));
+
     const model = new SentenTreeBuilder()
-      .buildModel(data);
+      .buildModel(this.data);
 
     this.vis = new SentenTreeVis(el)
       .data(model.getRenderedGraphs(graphs));
--- a/components/SentenTree/index.js
+++ b/components/SentenTree/index.js
@@ -26,6 +26,7 @@ export default class SentenTree extends VisComponent {
       {
         name: 'text',
         description: 'The field containing the text sample.',
+        type: 'string',
         domain: {
           mode: 'field',
           from: 'data',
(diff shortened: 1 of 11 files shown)


mgrauer:
Can you add a screenshot if it's easy?  I'm curious what this thing is.

🍛

ronichoudhury:
@mgrauer, sure thing, screencap added to top comment.

ronichoudhury:
Rebased on master.

jeffbaumes:
@ronichoudhury this. is. cool.

Looks like this has conflicts now, probably pretty simple to rebase.

ronichoudhury:
Rebased.

components/SentenTree/index.js
```diff
@@ -0,0 +1,57 @@
...
+        domain: {
+          mode: 'field',
+          from: 'data',
+          fieldTypes: ['integer']
+        }
+      }
+    ];
+  }
+
+  constructor (el, {data, graphs = 3}) {
```
jeffbaumes: It seems we are not using the `id`, `text`, and `count` parameters here. I assume we would transform the data to put those fields in standard places that SentenTree expects before initializing the visualization. It would also be great to put in defaults (set an `id` to a unique number 0 to N if no `id` is provided, set all counts to 1 if no `count` is provided). That way all that would be needed is a table that has a text column.

Also, the `graphs` parameter should be described as an option so there can be auto UI generation for setting the number of graphs to show.

ronichoudhury: Good ideas. Check out 3c69e3d and 59bec69. In particular, did I get `type` and `format` right for the `graphs` option?

jeffbaumes:
Looking good. I discovered that you are not emptying the `div` before rendering the component. Could you add that in?

ronichoudhury:
Oops. Added in a new commit.

It might be worthwhile to put some of this stuff (emptying top-level div, prepping datasets with column selectors) into helper functions. I'll file an issue to that effect.

components/SentenTree/index.js
```diff
@@ -0,0 +1,71 @@
...
+        type: 'string',
+        domain: {
+          mode: 'field',
+          from: 'data',
+          fieldTypes: ['string', 'integer', 'number']
+        }
+      },
+      {
+        name: 'text',
+        description: 'The field containing the text sample.',
```
jeffbaumes: needs `type: 'string'`

ronichoudhury:
Is it better to leave it as `integer` and fix upstream, or change it to `number` so that it works now?

jeffbaumes:
Since this is not on a critical path, I'd say leave as integer and support it as needed in downstream tools.

PR #491, opened 2017-05-04
Add GLO component

ronichoudhury:
TODO:
- [x] documentation
  - [x] Include link to GLOSTIX paper/website
  - [x] Add the same to docs for SentenTree
- [ ] ~~translation of "techniques" from GLO repository~~


ronichoudhury:
@jeffbaumes this is ready for review.

I decided not to add the "techniques" because we need some further refinement to the component before they will "just work".

ronichoudhury:
@jeffbaumes PTAL

webpack.js
```diff
@@ -59,14 +59,24 @@ module.exports = function (config, basePath, options) {
...
+      include: includePaths.concat(gloPath)
+    },
+    {
+      test: /\.js$/,
+      tag: 'glo',
+      loaders: [
+        'exports-loader?GLO',
+        'imports-loader?_=underscore&cola=webcola'
+      ],
+      include: gloPath
```
jeffbaumes: We were thinking of making GLO do this itself so that we didn't need this loader. I'm ok with a separate issue if you don't want to tackle this now.

config/webpack-examples.config.js
```diff
@@ -78,6 +78,14 @@ module.exports = {
...
       },
       {
+        test: /\.js$/,
+        loaders: [
+          'exports-loader?GLO',
+          'imports-loader?_=underscore&cola=webcola'
+        ],
+        include: __dirname + '/node_modules/glo/glo.js'
+      },
+      {
```
jeffbaumes: This loader would also not be needed.

ronichoudhury: Oh, right. I'll handle that in a separate PR, since this works nicely as is.

PR #493, opened 2017-05-16
Add basic test for TreeHeatmap

jeffbaumes:
First testing to see what this does to the coverage report.

### DIFF
--- a/components/TreeHeatmap/index.js
+++ b/components/TreeHeatmap/index.js
@@ -84,6 +84,7 @@ export default class TreeHeatmap extends VisComponent {
 
   constructor (el, options) {
     super(el);
+    options = options || {};
     this.data = options.data;
     this.scale = options.scale || 'global';
     this.clusterRows = options.clusterRows === undefined ? true : options.clusterRows;
@@ -98,6 +99,10 @@ export default class TreeHeatmap extends VisComponent {
 
   render () {
     d3.select(this.el).selectAll('*').remove();
+
+    if (this.data === undefined || this.data.length === 0) {
+      return;
+    }
 
     let size = getElementSize(this.el);
     let width = this.width || size.width;
@@ -119,16 +124,20 @@ export default class TreeHeatmap extends VisComponent {
 
     let keys = Object.keys(this.data[0]);
     let idColumn = this.idColumn;
-    if (!idColumn) {
+    if (idColumn === undefined) {
       idColumn = keys.includes('_id') ? '_id' : idColumn;
       idColumn = keys.includes('_') ? '_' : idColumn;
       idColumn = keys.includes('') ? '' : idColumn;
+    }
+    if (idColumn === undefined) {
+      console.log('TreeHeatmap: No suitable idColumn found.');
+      return;
     }
 
     let rows = [];
     let reachedMetadata = false;
     this.data.forEach(row => {
-      let id = row[idColumn];
+      let id = '' + row[idColumn];
       if (id === '_child1') {
         reachedMetadata = true;
       }
@@ -151,7 +160,7 @@ export default class TreeHeatmap extends VisComponent {
     let originalDataRows = {};
     reachedMetadata = false;
     this.data.forEach(row => {
-      let id = row[idColumn];
+      let id = '' + row[idColumn];
       if (id === '_child1') {
         reachedMetadata = true;
       }
@@ -272,7 +281,7 @@ export default class TreeHeatmap extends VisComponent {
       for (let link = 0; link < links.length; link += 1) {
         linkMap[links[link].cluster] = links[link];
       }
-      let finalCluster = links[links.length - 1];
+      let finalCluster = links[links.length - 1] || {};
       finalCluster.offset = 0;
       finalCluster.parent = finalCluster;
       for (let link = links.length - 1; link >= 0; link -= 1) {
--- a/components/TreeHeatmap/index.js
+++ b/components/TreeHeatmap/index.js
@@ -136,7 +136,7 @@ export default class TreeHeatmap extends VisComponent {
     let rows = [];
     let reachedMetadata = false;
     this.data.forEach(row => {
-      let id = String(row[idColumn]);
+      let id = '' + row[idColumn];
       if (id === '_child1') {
         reachedMetadata = true;
       }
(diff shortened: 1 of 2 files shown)


jeffbaumes:
Realized I took out the fancy animated transitions - I'll need to get those back in before we merge.

jeffbaumes:

components/TreeHeatmap/index.js
```diff
@@ -119,16 +124,20 @@ export default class TreeHeatmap extends VisComponent {
...
     let keys = Object.keys(this.data[0]);
     let idColumn = this.idColumn;
-    if (!idColumn) {
+    if (idColumn === undefined) {
       idColumn = keys.includes('_id') ? '_id' : idColumn;
       idColumn = keys.includes('_') ? '_' : idColumn;
       idColumn = keys.includes('') ? '' : idColumn;
     }
+    if (idColumn === undefined) {
+      console.log('TreeHeatmap: No suitable idColumn found.');
```
ronichoudhury: Is this a fatal error? If so, instead of using `console.log()`, we should throw an `Error` object with the error message. Even if it's not, I'd like to not rely on console log statements to report warning conditions.

components/TreeHeatmap/index.js
```diff
@@ -119,16 +124,20 @@ export default class TreeHeatmap extends VisComponent {
...
+      return;
+    }
 
     let rows = [];
     let reachedMetadata = false;
     this.data.forEach(row => {
-      let id = row[idColumn];
+      let id = '' + row[idColumn];
```
ronichoudhury: Is this meant to cast `row[idColumn]` to a string? Is there a better way to do it, perhaps with the `String()` constructor?

components/TreeHeatmap/index.js
```diff
@@ -151,7 +160,7 @@ export default class TreeHeatmap extends VisComponent {
     let originalDataRows = {};
     reachedMetadata = false;
     this.data.forEach(row => {
-      let id = row[idColumn];
+      let id = '' + row[idColumn];
```
ronichoudhury: See above comment.

components/TreeHeatmap/index.js
```diff
@@ -177,7 +186,7 @@ export default class TreeHeatmap extends VisComponent {
     let colLinks = [];
     if (clusterColumns) {
       columns.forEach(col => {
-        if (+metadataRows._cluster[col.name] >= 0) {
+        if (metadataRows._cluster && +metadataRows._cluster[col.name] >= 0) {
```
ronichoudhury: Similar thing here as for cast-to-string - I think best practice now is to use `Number()` instead of the unary `+` (but [citation needed]).

components/TreeHeatmap/test/treeheatmap.js
```diff
@@ -0,0 +1,46 @@
+import test from 'tape-catch';
+import TreeHeatmap from '..';
+
+test('TreeHeatmap component ', t => {
```
ronichoudhury: Remove trailing space from test name.

components/TreeHeatmap/test/treeheatmap.js
```diff
@@ -0,0 +1,46 @@
...
+  vis.render();
+  t.equal(el.childNodes.length, 0, 'no idColumn should do nothing');
+
+  vis = new TreeHeatmap(el, {data: [{_id: 0, a: 0}]});
+  vis.render();
+  t.equal(el.childNodes.length, 1, 'single row and column should produce something');
+  console.log(el.innerHTML);
+  let single = document.createElement('div');
+  single.innerHTML = '<svg width="400px" height="400px"><clipPath id="clip-rect"><rect x="100" y="100" width="200" height="200"></rect></clipPath><g clip-path="url(#clip-rect)"><rect class="datum" fill="rgb(255, 255, 255)" x="100" y="100" opacity="1" width="200" height="200"></rect></g><clipPath id="clip-row-labels"><rect x="300" y="100" width="100" height="200"></rect></clipPath><g clip-path="url(#clip-row-labels)"><text class="row-label" color="black" font-size="10px" alignment-baseline="middle" x="300" y="200">0</text></g><clipPath id="clip-col-labels"><rect x="100" y="300" width="200" height="100"></rect></clipPath><g clip-path="url(#clip-col-labels)"><g transform="translate(200,300)" class="col-label"><text color="black" font-size="10px" transform="rotate(-90)" text-anchor="end" alignment-baseline="middle">a</text></g></g></svg>';
```
ronichoudhury: Apparently with ES6 string templates, you can freely mix newlines into the string and they will show up literally. This line would read a lot easier if we tried that here.

components/TreeHeatmap/index.js
manthey: I'm going to support the `'' + (value)` to cast to string technique.  `String()` may be clearer to some, but it is 15x slower (https://jsperf.com/cast-to-string/2), so it is a bad practice to get into.  Empty string concatenation has been the de-facto string cast for long enough to be clear.

components/TreeHeatmap/index.js
jeffbaumes: This is a condition where nothing can be rendered since there is no ID column specified, so yes I think this is fatal for initializing the visualization. I'll modify this to a throw and I think I can change the test to use `t.throws` to check for it.

jeffbaumes:
@ronichoudhury I think I addressed all comments, PTAL.

components/TreeHeatmap/index.js
ronichoudhury: That's so weird - do you know why `String(x)` can't just do whatever `'' + x` does under the hood?

In any case, @jeffbaumes, maybe we should revert this particular change in light of @manthey's performance concerns.

manthey: I don't really know (see http://stackoverflow.com/questions/14918778/why-concat-empty-character-is-faster-than-tostring).

I view using `'' + x` for strings and `+x` for numbers as similar syntax to using `=>` for functions rather than `function` and binding-- it is not obvious to the novice, but is what the language has evolved to use and is common.

jeffbaumes: Thanks @manthey, a revert is on the way

jeffbaumes:
Ok @ronichoudhury, may be ready for a final check.

PR #500, opened 2017-05-30
Remove Resonant Lab

ronichoudhury:
The code (with history) was moved to the Resonant Lab reimplementation project; see https://github.com/ronichoudhury-work/resonantlab/pull/68.


ronichoudhury:
Will do!

PR #501, opened 2017-05-31
Fix TrackerDash example and add a test

zackgalbreath:


ronichoudhury:
@zackgalbreath I pushed a couple of commits to fix some minor issues surrounding d3 v3. PTAL and make sure your changes are still good.

zackgalbreath:
@ronichoudhury I force-pushed this branch to address your feedback, PTAL.  Also note the CircleCI test failures.  Are these expected?

ronichoudhury:
Thanks @zackgalbreath.

Both of those test failures are fixed in the branch I'm working on, but in a way that it's difficult to backport to your branch. I'll go ahead and manually bring those patches over to this branch, stand by.

ronichoudhury:
Ok, I patched the file that I think is causing the problem - let's see what happens now.

ronichoudhury:
@zackgalbreath I had to re-run without caching but it looks like things are ok now :tada:

I'll approve the PR but please take one more look to make sure the code looks ok after my fixes.

Thanks!

components/TrackerDash/ResultTablePane.js
```diff
@@ -1,6 +1,6 @@
 import _ from 'underscore';
 import $ from 'jquery';
-import * as d3 from 'd3';
+import { median } from 'd3-array';
```
zackgalbreath: I feel a bit uneasy about mixing d3v3 and d3v4 in TrackerDash.  Is there any reason why we shouldn't consistently use d3v3 for this component?

ronichoudhury: Oops, fixing...

ronichoudhury: See e336b2a97e6778cffb6413f762ba648561046581.

PR #502, opened 2017-06-05
[WIP] Add rudimentary unit test for TrackerDash

zackgalbreath:
This test currently only verifies that the TrackerDash consists of a single parent element.

### DIFF
--- a/app/examples/trackerdash/index.js
+++ b/app/examples/trackerdash/index.js
@@ -368,3 +368,5 @@ if (qParams) {
 }
 var app = new TrackerDash(document.body, appSettings);
 app.render();
+
+export default appSettings;
--- a/components/TrackerDash/test/trackerdash.js
+++ b/components/TrackerDash/test/trackerdash.js
@@ -1,0 +1,11 @@
+import test from 'tape-catch';
+import TrackerDash from '..';
+import appSettings from '../../../app/examples/trackerdash';
+
+test('TrackerDash component', t => {
+  let el = document.createElement('div');
+  let vis = new TrackerDash(el, appSettings);
+  vis.render();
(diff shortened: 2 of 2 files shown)


zackgalbreath:
This buys us [a pretty good bump in coverage](https://codecov.io/gh/Kitware/candela/compare/master...trackerdash_unit_test/changes), but this test (as currently written) doesn't do much to protect the behavior of this component.

Thoughts on how to proceed? As I see it, we have two options here:

1) Instead of instantiating TrackerDash directly, build up coverage by testing its individual sub-components instead.

2) Accept this big bump of coverage and gradually try to add more predictive test cases for TrackerDash, even though they won't result in any additional coverage.

jeffbaumes:
Instead of performing a full test of each component, we could dig in to a few of the sub-components in the TrackerDash DOM with this test. To help the process, you could inspect the TrackerDash example to see what the HTML looks like, then update the test to examine some of the children (e.g. are the right number of bullet charts created? is the summary plot there? do some of the data-driven headings have the correct text?)

ronichoudhury:
@zackgalbreath, what do you think of @jeffbaumes's suggestion?

The changes in this PR look good to me, but we have to decide how to go forward with the testing strategy itself.

mgrauer:
I'd favor setting up the testing skeleton with one or two pieces of meat, then saving the rest of the fill in for later.  Once the skeleton is in place, increasing coverage could be a good learning opportunity for someone new to Candela.

ronichoudhury:
@zackgalbreath, should we go ahead and merge this now, and decide how to proceed more fully later?

Sorry to revive an old thread but we should definitely do something with this PR :)

PR #509, opened 2017-07-12
refactor(vega): Move to Vega 3.0, VegaLite 2.0

jeffbaumes:
* Remove JSON templating that was used for Vega 2.x specs.
* Make a new VegaView superclass that utilizes a generateSpec method
  instead of a JSON spec.
* Remove BulletChart, Heatmap which had limited utility.
* SurvivalPlot is taken out of the build for now until we decide
  what to do with it.
* [x] Figure out what to do with SurvivalPlot (remove for now).
* [x] Fix tests.
* [x] Fix examples.
* [x] Fix dynamic line chart
* [x] Fix resize examples.
* [x] Update documentation.


jeffbaumes:

ronichoudhury:
There are still a few "regressions" (not reflected in the tests) caused by this PR, but I think we should merge this now and fix the issues in future topics. Here's a list:
- #511: shapes don't appear in `scatter` example (but do in `scatter-solid`)
- #512: the axis ranges do not update naturally in the `line-dynamic` example
- #513: the `line-points` example doesn't work anymore, because the new `LineChart` doesn't listen for click events like the old one did

.eslintrc.json
```diff
@@ -0,0 +1,3 @@
+{
+  "extends": "semistandard"
+}
```
ronichoudhury: Why do we need this file?

ronichoudhury:
@jeffbaumes since you're away, I'll press the button. Hope you don't mind :)

.eslintrc.json
jeffbaumes: Not exactly sure, but it makes my Atom editor eslint plugin happy. Cool kids like girder have one :)

PR #516, opened 2017-08-04
feat(LineChart): Add option for setting x and y scales

### LINKED ISSUE #512: "line-dynamic" example behaves weirdly

jeffbaumes:
Fixes #512


jeffbaumes:
@ronichoudhury take a look. I've added a way to set scale options, that way you can set `nice` and `zero` to false, which makes the dynamic line plot look much nicer.

ronichoudhury:
Fantastic!

PR #540, opened 2018-05-16
fix(vega): Remove unneeded deps to avoid version/interaction issues

jeffbaumes:
Without this, some incompatible versions of vega-* libraries were pulled in, causing interaction to break.

### DIFF
--- a/packages/vega/package.json
+++ b/packages/vega/package.json
@@ -31,9 +31,7 @@
     "@candela/core": "^0.21.0",
     "d3-selection": "^1.1.0",
     "vega": "^3.0.0-rc1",
-    "vega-lite": "^2.0.0-beta.8",
-    "vega-schema-url-parser": "^1.0.0-beta.2",
-    "vega-tooltip": "^0.4.2"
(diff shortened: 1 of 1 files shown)


jeffbaumes:
@waxlamp ok this PR turned into a no-op. It seems that doing the following fixed things up for me. Please verify and I'll close this one.

```
lerna clean
rm -rf ./node_modules
rm package-lock.json

npm install
npm run build
npm run build:examples
npm run examples
```

waxlamp:
This sequence worked for me -- I have an interactive scatter plot again.

jeffbaumes:
There is still a slight bit of mystery because the published @candela/all v0.21.0 still appears to have the non-interaction issue. There might be some tweaking needed in circle's yaml? For example, when does the "npm install" step happen there? Or maybe the problem will just "go away" on the next version?

waxlamp:
@jeffbaumes, do you think this is still a problem? If there's a way to reproduce it, could you open an issue describing how to do so?

PR #543, opened 2018-05-23
Add `calculateSize` option to VegaView

waxlamp:
This prevents by default `VegaView`'s behavior of rerendering multiple times to figure out what the correct size should be for small multiple views. The behavior is still available for use when needed (e.g., with `ScatterPlotMatrix`).


waxlamp:
@jeffbaumes, the tests are passing now, but I rebased this branch on #544, so both branches will need a review approval before I can merge them.

Thanks!

PR #545, opened 2018-06-04
Use vega-lib in place of vega

waxlamp:


waxlamp:
@jeffbaumes: once this is merged, I can work on updating the Girder Candela plugin :smile:

PR #546, opened 2018-06-05
Include babel-loader in all test configurations

waxlamp:


jeffbaumes:
What's the context of this change, what problem does it solve?

waxlamp:
There's no evident problem that it solves; for some reason, without this loader the tests seem to build properly, but for completeness I wanted to include the loader rule.

waxlamp:
@jeffbaumes could you take another look at this please (made the change I discussed with your yesterday, which caused your earlier approval to be dismissed).

jeffbaumes:
Circle is complaining with:

```
TypeError: Class constructor ScatterPlot cannot be invoked without 'new' FAILED
```

Is there something different about the transpilation that is affecting this?

waxlamp:
I rebased this branch on master to properly incorporate the build improvements there--and if this doesn't just work, I'm inclined to drop this PR.

waxlamp:
@jeffbaumes, looks like the problem is fixed now.

